PR #3, opened 2020-02-12
Minor readme update

ngaranko:
Adding info on what to do in order to import schemas


vdboor:
nice! merged with rebase

PR #14, opened 2020-02-20
Limit fields using query parameter

ngaranko:


src/tests/test_rest_framework_dso/test_serializers.py
```diff
@@ -72,3 +73,45 @@ def test_pagination_many(api_rf, movie):
...
+
+@pytest.mark.django_db
+def test_fields_limit_works(api_rf, movie):
+    """Prove that serializer can limit output fields."""
+    django_request = api_rf.get("/", {"fields": "name"})
+    request = Request(django_request)
+    queryset = Movie.objects.all()
+
+    serializer = MovieSerializer(many=True, instance=queryset)
+    serializer.context["request"] = request
```
vdboor: Small catch: couldn't you call `MovieSerializer(many=True, instance=queryset, context={"request": request})`?

src/rest_framework_dso/serializers.py
```diff
@@ -156,6 +172,23 @@ def many_init(cls, *args, **kwargs):
     def to_representation(self, instance):
         ret = super().to_representation(instance)
 
+        display_fields = self.get_fields_to_display()
+        if display_fields is not None:
+            if not (set(display_fields) <= set(ret.keys())):
+                # Some of `display_fields` are not in result.
+                invalid_fields = [f for f in display_fields if f not in ret.keys()]
```
vdboor: `f not in ret.keys()` can be `f not in ret` as dict.__contains__ checks for the key :-)

vdboor:
Changes are in new branch: `feature/limit_fields`

PR #47, opened 2020-05-26
Range filtering.

ngaranko:
This PR allows parsing of `additionalFilters` schema parameter and transforms it into new additional filters that are mapped to one or multiple fields.

Including `range` filter that will create Range filter `start__lte=x` & `end__gt=x`.

Schema will look as follows:

```
"tables": {
   "x":  {
      "schema":  {
        ...
        "additionalFilters": {
          "regimes.inWerkingOp": {                                     << String to be presented in URI.
            "type": "range",                             << Filter name. For now we have only `range`.
            "start": "regimes.begin tijd",          << Start field name. Relations are separated with dots.
            "end": "regimes.eind tijd"              << End field name. Relations are separated with dots.
          }
        },
        "properties": {...}
     }
```
Full example: https://github.com/Amsterdam/dso-api/pull/47/files#diff-b4fad29bf58070e15c0606a6475b2f87

### DIFF
--- a/src/rest_framework_dso/filters.py
+++ b/src/rest_framework_dso/filters.py
@@ -20,11 +20,13 @@ from django.contrib.postgres.fields.array import ArrayField
 from rest_framework.exceptions import ValidationError
 from rest_framework.filters import OrderingFilter
 from rest_framework_gis.filters import GeometryFilter
+from schematools.types import slugify
 
 __all__ = [
     "DSOFilterSet",
     "DSOFilterSetBackend",
     "DSOOrderingFilter",
+    "EffectiveFilter",
 ]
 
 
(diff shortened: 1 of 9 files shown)


src/rest_framework_dso/filters.py
```diff
@@ -73,6 +75,32 @@ def __init__(self, field_name=None, lookup_expr="exact", **kwargs):
...
+        return qs.filter(
+            **{f"{self.start_field}__lte": value, f"{self.end_field}__gt": value}
+        )
+
+    def convert_field_name(self, field_name):
+        if "." in field_name:
+            return "__".join(
+                [self.convert_field_name(part) for part in field_name.split(".")]
+            )
+        return slugify(field_name, sign="_")
```
jjmurre: Let op, 'sign' wordt niet meer ondersteund in nieuwe versie van string-utils. Is nu 'separator'.

PR #78, opened 2020-07-08
Chris/display

cvriel:
[DS-325]  if not present/empty display field, content in "title" is omitted


cvriel:
Redeploy nodig

PR #79, opened 2020-07-08
Chris/displayfield

cvriel:
[DS-325] If display field is present then show else omit in outfield element title


cvriel:
Redeploy nodig

cvriel:
Redeploy nodig

PR #81, opened 2020-07-08
Chris/displayfield

cvriel:
[DS-325] Content of display field is shown in title element if present. If the display field is not present or empty the title element omitted.


cvriel:
Close for redeploy

PR #105, opened 2020-08-27
Format 500 (and other) errors according to DSO-API standards

bheupers:
Use X-Unique-ID passed by HAProxy as instance value


cvriel (approved):
Netjes!

PR #112, opened 2020-09-14
[DS-428] Allow multiple [not] filters to exclude multiple items

vdboor:


jjmurre (approved):
Mooi, ziet er goed uit!

PR #127, opened 2020-10-20
Coordinates from the DSO api are served as lng-lat

timvanoostrom:


ngaranko (approved):

ngaranko:
Bedankt @timvanoostrom !

PR #152, opened 2020-12-10
Do not retrieve the primary key of the through table, name may be dif…

vandoornik:
Removed accidental dependency on the PK of the through table.


cvriel (approved):

jjmurre (approved):
LGTM

PR #157, opened 2020-12-21
overriding serializer_field_mapping not globally but in class constru…

vandoornik:
…ctor

A small refactoring to avoid monkey-patching the django rest framework.


jjmurre (approved):
Inderdaad netter om het hier te doen!

PR #166, opened 2021-01-13
GeoJSON export format for API vieews

vdboor:
* Upgraded orjson library which support serializing subclasses of dict such as GeoJsonDict that GeometryField provides.
* embedded content is not yet supported (will be later when making the API streaming)
* Pagination is delegated to the renderer so it can be written in the correct format.


ngaranko (approved):
LGTM

src/dso_api/dynamic_api/views/api.py
```diff
@@ -129,10 +127,10 @@ class DynamicApiViewSet(
     @property
     def paginator(self):
         if (
-            self.request.accepted_renderer.format == "csv"
+            self.request.accepted_renderer.format in ("csv", "geojson")
             and self.pagination_class.page_size_query_param not in self.request.GET
         ):
-            # Avoid expensive COUNT(*) for CSV formats,
+            # Avoid expensive COUNT(*) for CSV and GeoJSON formats,
```
cvriel, line 133: Top

PR #175, opened 2021-02-01
Feature/small fixes

vdboor:
PR that fixes:

* middleware usage (process_request will no longer be magically called since Django 1.11+).
* better definition of WGS84 in GeoJSON
* move unlimited page sizes decision to rendering classes


jjmurre (approved):
LGTM

PR #179, opened 2021-02-03
Fix content-type on streaming responses, which is determined too late

vdboor:
The response headers were already determined before reading the streaming
content.


ngaranko (approved):
LGTM

PR #180, opened 2021-02-03
Optimize loose M2M field collection, called during each loop

vdboor:
This avoids finding out which fields are loose M2M relations each loop.
When a full export is done (e.g. 50.000 items), this also called `toCamelCase()` 100.000 times.


vandoornik (approved):
Nice improvement. Perhaps _loose_relation_m2m_fields or part of it can be moved to M2MRelationMaker in schematools/contrib/django/factories.py

jjmurre (approved):
Caching to the rescue, LGTM.

PR #182, opened 2021-02-03
Massive performance fix of serializer.fields access.

vdboor:
The original DRF `Serializer.fields` is a cached_property that reads get_fields() once.
Thus overriding `get_fields()` instead of `fields` preserves this caching.

Currently the fields were read multiple times per object, while doing massive unauthorized-fields checks.
This change fixes that.

To address any concerns about information-leakage serializers *must* receive an request object so the unauthorized field checks are performed. This only conflicted with the OAS schema view, which is fixed by `public=False`. Assigning request.auth_profile had to be moved because of this.

The CSV export can not also avoid reading all fields (including M2M), while not returning these.

### DIFF
--- a/src/dso_api/dynamic_api/middleware.py
+++ b/src/dso_api/dynamic_api/middleware.py
@@ -3,6 +3,7 @@ import logging
 
 from django.http import UnreadablePostError
 from django.utils.deprecation import MiddlewareMixin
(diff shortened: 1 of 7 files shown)


ngaranko (approved):
LGTM.

PR #191, opened 2021-02-15
Use generic CMDB deployment

daveneijsen:

Change deployment to use the generic deploy.yml in Openstack, which uses the CMDB deployment mechanism.


daveneijsen:
Dennis says: LGTM

PR #193, opened 2021-02-17
REST API streaming

vdboor:
* One commit to introduce streaming rendering
* One commit to implement iterator-based data retrieval
* Various bugfixes (as separate commits)

The core machinery is the `ObservableIterator` that fills the `EmbeddedResultSet` object with all ID's that need to be fetched. The `get_related_list_ids()` is obsolete and removed, everything happens on the instance directly.

Response changes:
* In the detail view, M2M fields always render as array `[]`. Previously the detail page would show a single M2M item as object. This was an oversight of the commits that introduced M2M relations (673e2571b48a3f69e5ad3f92e760f950843d496e).
* The list page _embedded always uses `[]` for missing results, instead of `null`.
* The browsable API uses 2 spaces for indentation.


src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,196 @@
...
+        self._has_items = True
+
+        # Notify observers
+        for notify_callback in self._item_callbacks:
+            notify_callback(value)
+
+        return value
+
+    def __len__(self):
+        if not self.is_iterated:
```
lbam, line 114: `AttributeError: 'IteratorObserver' object has no attribute 'is_iterated'`

Is deze methode überhaupt nodig?

vdboor: ok, fixed.. Dit was vooral bedoelt om pagination makkelijker te maken.

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,197 @@
...
+        return value
+
+    def __len__(self):
+        if not self._is_iterated:
+            raise RuntimeError("Unable to read counting generator size yet")
+        return self.number_returned
+
+    def __bool__(self):
+        """Tell whether the generator would contain items."""
+        if self._has_items is None:
```
lbam, line 121: AttributeError op _has_items

vdboor: Tx, fixed.. arg, teveel heen en weer geschoven met code

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # Notify observers
+        for notify_callback in self._item_callbacks:
+            notify_callback(value)
+
+        return value
+
+    def __len__(self):
+        if not self._is_iterated:
+            raise RuntimeError("Unable to read counting generator size yet")
```
lbam, line 117: Beetje een terzijde, maar dit maakt bijv.

```python
it = IteratorObserver([1,2,3])
list(it)
```

onmogelijk, want list krijgt de RuntimeError. Zonder __len__ werkt dat wel. Dit is altijd het eerste wat ik doe als ik een iterator moet debuggen.

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,198 @@
...
+    if not available:
+        return ParseError("Eager loading is not supported for this endpoint")
+    else:
+        return ParseError(
+            f"Eager loading is not supported for field '{field_name}', "
+            f"available options are: {available}"
+        )
+
+
+class IteratorObserver:
```
lbam, line 77: Nitpick: is het niet een ObservableIterator? De observers zijn de dingen die je aan de constructor meegeeft, de iterator is wat die oplevert.

src/rest_framework_dso/embedding.py
vdboor: Removed!

src/rest_framework_dso/embedding.py
vdboor: ok, goede debugging 👍  Dit werkt idd niet, methode is weggehaald.

src/rest_framework_dso/embedding.py
vdboor: Daar heb je helemaal gelijk in. De naam `IteratorObserver` is vanaf het perspectief van de gebruiker wel oké, maar het object zelf is inderdaad een observable.

vdboor: Ook direct tests toegevoegd :-)

src/tests/test_rest_framework_dso/test_embedding.py
```diff
@@ -0,0 +1,39 @@
...
+        """Prove that the final outcome is as desired"""
+        seen1 = []
+        seen2 = []
+
+        observer = ObservableIterator("abcd", observers=[seen1.append, seen2.append])
+        assert list(observer) == ["a", "b", "c", "d"]
+
+        assert seen1 == ["a", "b", "c", "d"]
+        assert seen1 == seen2
+        assert bool(observer)
```
lbam, line 15: Dit beantwoordt ook mijn vraag of `__bool__` nog steeds True moet geven als alle items geconsumeerd zijn :+1:

lbam:
Gedeelte dat ik kan beoordelen ziet er goed uit.

src/tests/test_rest_framework_dso/test_embedding.py
vdboor: 😄

jjmurre (changes requested):
Ziet er prima uit! Ben benieuwd naar de performance. Ik kom nog 1 raar bugje tegen. In dit endpoint krijgt je invalide json `/v1/bag/ligplaatsen/?_expandScope=heeftDossier`. Alle andere te expanden velden gaan goed, maar deze geeft:

```Error: Parse error on line 34:
...","gebruiksdoel":[]}
-----------------------^
Expecting ',', ']'
```
Als je opvraagt met `_format=json`.

src/rest_framework_dso/serializers.py
```diff
@@ -93,6 +93,13 @@ def to_representation(self, data):
             # Trick DRF into generating the response bit by bit, without
             # caching the whole queryset into memory.
             items = (self.child.to_representation(item) for item in data.iterator())
+
+            # Make sure the first to_representation() is called early on. This makes sure
+            # DSOModelSerializer.to_representation() can inspect the CRS, and define the
+            # HTTP header for the Content-Crs setting. This will be too late when the streaming
+            # rendering started. As bonus, it's a good guard against database errors, since those
+            # exceptions might not be rendered nicely rendered either once streaming started.
```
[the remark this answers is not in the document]
vdboor: fixed

src/dso_api/dynamic_api/serializers.py
```diff
@@ -54,35 +54,27 @@ def to_representation(self, fields_to_be_encoded: list, data):
...
-        table_name = model._meta.db_table
-        fetcher = partial(_fetch_by_ids, identifier, sequence_name, table_name)
+    def _fetcher(id_list):
+        if is_loose:
+            # We assume temporal config is available in the dataset
+            dataset = model.get_dataset()
+            identifier = dataset.identifier
+            sequence_name = dataset.temporal["identifier"]
+            return (
+                model.objects.distinct(identifier)  # does SELECT DISTINCT ON(identifier)
```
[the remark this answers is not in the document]
vdboor: 😄

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,192 @@
...
+
+from dso_api.dynamic_api.permissions import fetch_scopes_for_model
+from rest_framework_dso.fields import AbstractEmbeddedField, EmbeddedField
+from rest_framework_dso.serializer_helpers import ReturnGenerator, peek_iterable
+
+EmbeddedFieldDict = Dict[str, AbstractEmbeddedField]
+
+
+def get_expanded_fields(  # noqa: C901
+    parent_serializer: serializers.Serializer, fields_to_expand: Union[list, bool]
```
gkoller, line 20: The list can be make more specific:
```python
fields_to_expand: Union[List[str], bool]
```
Assuming the fiield names are of type string.

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,192 @@
...
+        except AttributeError:
+            raise RuntimeError(
+                f"{parent_serializer.__class__.__name__}.{field_name}"
+                f" does not refer to an embedded field."
+            ) from None
+
+        # Check access via scopes (NOTE: this is higher-level functionality)
+        scopes = fetch_scopes_for_model(field.related_model)
+        if auth_checker and not auth_checker(*scopes.table):
+            # Not allowed, silently drop for _expand=true request.
```
gkoller, line 53: Don't know what kind of logging we do in DSO-API, but things like this generally do warrant a logging statement.

src/rest_framework_dso/embedding.py
```diff
@@ -0,0 +1,192 @@
...
+    if not available:
+        return ParseError("Eager loading is not supported for this endpoint")
+    else:
+        return ParseError(
+            f"Eager loading is not supported for field '{field_name}', "
+            f"available options are: {available}"
+        )
+
+
+class ObservableIterator:
```
gkoller, line 77: Brand new class. Hence could do with some type annotations. Clearly from the name it suggests its an `Iterable`, but I'm not sure yet what it iterates over. Something like this would help:
```
class ObservableIterator(Iterator[Fubar]):
```
Where `Fubar` is replaced by whatever it is `ObservableIterator` iterates over.

This also means that you should probably type annotate the member functions as well.

```python
def __iter__(self) -> Iterator[Fubar]:
   ...

def __next__(self) -> Fubar:
   ...
```

src/rest_framework_dso/renderers.py
```diff
@@ -82,6 +107,61 @@ class HALJSONRenderer(RendererMixin, renderers.JSONRenderer):
...
+
+        return orjson.dumps(data, option=orjson.OPT_INDENT_2)
+
+    def _render_json(self, data, level=0):
+        if not data:
+            yield orjson.dumps(data)
+            return
+        elif isinstance(data, dict):
+            # Streaming output per key, which can be a whole list in case of embedding
+            yield b"{"
```
gkoller, line 139: Instead of using bytes directly you might want to consider using `.encode()` to convert `str`s to `bytes`. That is arguably more readable.

src/rest_framework_dso/serializer_helpers.py
```diff
@@ -23,3 +25,16 @@ def __repr__(self):
     def __bool__(self):
         # Generators are always true, calls to "if data: ..." should not fail.
         return True
+
+
+def peek_iterable(generator):
```
gkoller, line 30: I'm a big fan of `more-itertools`. It has functions for looking ahead and back:

https://more-itertools.readthedocs.io/en/stable/api.html?highlight=peek#lookahead-and-lookback

Might want to consider using that.

src/tests/test_rest_framework_dso/test_serializers.py
```diff
@@ -69,21 +74,25 @@ def test_serializer_embed_with_missing_relations(drf_request):
...
+
+        data = normalize_data(serializer.data)
+        assert data == {
+            "movie": [{"name": "Test", "category_id": movie.category_id}],
+            "category": [],
+        }
+    finally:
+        # Make sure the cleanup happens or unit tests report this error
+        # instead of the actual assertion error.
+        cursor.execute("DELETE FROM test_rest_framework_dso_movie WHERE category_id=333")
```
gkoller, line 95: Why is this cleanup necessary? I would have expected `@pytest.mark.django_db` to rollback the transaction at the end of this test function automatically?

src/rest_framework_dso/renderers.py
gkoller: On second thought, `b""` is just fine now that I've found out orjson returns bytes instead of `str` (that's diffferent than what the stdlib json lib does)

Had orjson return str I  would have advocated for doing regular string formatting with a `.encode()` at the end (including for the literals as here). But as orjson returns bytes, using bytes for the literals is better here.

ngaranko (approved):
LGTM.

src/rest_framework_dso/response.py
```diff
@@ -76,7 +76,20 @@ def _read_rendered_content(self):
                 f"No stream generated by {self.accepted_renderer.__class__.__name__}"
             )
 
-        yield from stream
+        try:
+            yield from stream
+        except Exception as e:
+            yield self.render_exception(e)
```
gkoller, line 82: Not sure if this provides enough info. For the end user nor for us developers (when `settings.DEBUG == True`).
You should probably use [sys.exc_info()](https://docs.python.org/3.8/library/sys.html#sys.exc_info) here. Something along the lines of (untested):

```python
yield self.render_exception(*sys.exc_info())  # render_exception()'s signature should change for this to work)
```
`render_exception()` should be changed to something like (untested):
```python
def render_exception(self, type: Type[Exception], value: Exception, tb: Any) -> str:
    """Inform clients that the stream was interrupted by an exception."""
    if settings.DEBUG:
        return '"\n\n" {stack_trace}'.format(
            stacktrace="".join(traceback.format_exception(type, value, tb)))
    else:
        return '"\n\nRendering aborted due to: {exc_msg}'.format(
            exc_msg="".join(traceback.format_exception_only(type, value)))
```

vdboor: This method only exists to infor the client that the stream processing was interrupted with an exception. Purposefully, not much information is given, so avoid informing clients with internal server details. The actual exception is still raised and logged server-side.

src/tests/test_rest_framework_dso/test_serializers.py
vdboor: This code already existed before, only without a try..finally. At the top of the test, an invalid `category_id` is inserted. When resetting the database between tests, this causes a FK constraint error - which distracts from the actual test failure.

src/rest_framework_dso/serializer_helpers.py
vdboor: This is such small case, that I avoided bringing more-itertools in. Otherwise, yes, it would have been nice.

src/rest_framework_dso/embedding.py
vdboor: Done!

src/rest_framework_dso/embedding.py
vdboor: Done!

PR #196, opened 2021-02-18
Do not give key error for missing geofield if is not required.

bheupers:
In the production version for HaalCentraal brk the begrenzingPerceel is
missing. This now gives the following error:

https://sentry.data.amsterdam.nl/sentry/datapunt-apis-prod/issues/2680654/


ngaranko (approved):
LGTM

PR #197, opened 2021-02-18
Fixes to API docs (docs/ dir)

lbam:
* Omit default page size

Doesn't match the implementation, and is prone to change anyway.

* Explain wildcard syntax

* Warn about trailing slash when using QGIS

Omitting the trailing slash causes QGIS to not work, only emitting
cryptic warning messages in its OAFIP log (whatever that may be).

* Fixed links in temporele datasets

bagh/buurt become gebieden/buurten.

### DIFF
--- a/docs/source/generic/rest.rst
+++ b/docs/source/generic/rest.rst
@@ -42,9 +42,8 @@ altijd een JSON response geforceerd.
 Paginering
 ----------
 
-De resultaten hebben een standaard paginagrootte van 1000 records.
-
-De paginagrootte kan aangepast worden door een :samp:`&page_size={n}` query parameter toe te voegen aan de request URL.
+De resultaten worden gepagineerd teruggegeven.
+De paginagrootte kan aangepast worden door een :samp:`?page_size={n}` query parameter toe te voegen aan de request URL.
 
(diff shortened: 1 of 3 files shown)


docs/source/generic/rest.rst
```diff
@@ -42,9 +42,8 @@ altijd een JSON response geforceerd.
 Paginering
 ----------
 
-De resultaten hebben een standaard paginagrootte van 1000 records.
-
-De paginagrootte kan aangepast worden door een :samp:`&page_size={n}` query parameter toe te voegen aan de request URL.
+De resultaten worden gepagineerd teruggegeven.
+De paginagrootte kan aangepast worden door een :samp:`?page_size={n}` query parameter toe te voegen aan de request URL.
```
vdboor, line 46: Oops.. deze parameter mag zelfs omgezet worden naar `?_pageSize={n}`. De `page_size` is de oude parameter die we voor backwords compatibiliteit nog laten werken.

vdboor (approved):

docs/source/generic/rest.rst
lbam: Done

PR #203, opened 2021-02-23
Fix exception rendering for invalid _expand, due to gettext_lazy()

vdboor:
Fix sentry message for `_expand=fieldname`.

The orjson library doesn't render gettext_lazy() messages. While it's possible to insert a _default hook, the source is well-defined and fixable.

* Tests added.
* A related set of tests is indented for grouping.

### DIFF
--- a/src/rest_framework_dso/views.py
+++ b/src/rest_framework_dso/views.py
@@ -143,8 +143,8 @@ def exception_handler(exc, context):
         # our own API's would return these.
         normalized_fields = {
             "type": f"urn:apiexception:{exc.code}",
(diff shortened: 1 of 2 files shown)


ngaranko (approved):
LGTM.

That black thing does black things all the time.

vdboor:
> That black thing does black things all the time.

No, I changed the indenting of a set tests, because I grouped them in a `TestExpand` class.

jjmurre (commented):
Ziet er goed uit. Blijkbaar komen snelheidsmonsters als orjson wel met een prijs :-)

PR #206, opened 2021-02-24
Add string filterset to URLs

lbam:
This patch adds the isnull, not, isempty and like filters to URL fields.

Fixes DS-501, which asks for these fields, except like. I could have dropped like, but it comes for free with the string filterset and it can be useful for finding, e.g., non-HTTPS URLs (`?url[like]=http:*`).

Has a unit test and works with a local copy of the bekendmakingen dataset.

### DIFF
--- a/src/tests/test_rest_framework_dso/test_filters.py
+++ b/src/tests/test_rest_framework_dso/test_filters.py
@@ -42,6 +42,7 @@ class TestDSOFilterSet:
                 "name": ["exact"],
                 "category_id": ["exact", "in", "not"],
                 "date_added": ["exact", "lt", "lte", "gt", "gte", "not"],
+                "url": ["exact", "isnull", "not", "isempty", "like"],
             }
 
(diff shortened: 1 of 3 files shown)


src/tests/test_rest_framework_dso/test_filters.py
```diff
@@ -42,6 +42,7 @@ class Meta:
                 "name": ["exact"],
                 "category_id": ["exact", "in", "not"],
                 "date_added": ["exact", "lt", "lte", "gt", "gte", "not"],
+                "url": ["exact", "isnull", "not", "isempty", "like"],
```
ngaranko, line 45: `date_added` and `category_id` have extra tests, maybe good to add same for `url`?

lbam: Done, in existing test_filter_logic.

ngaranko (approved):

PR #211, opened 2021-03-01
FIX using camelCase table names vs feature names in WFS

cvriel:
The use of underscore - as a result of camelCase table names - was not taken into account when building the WFS viewset.

This led to a 400 Bad Request error. Which was a result of not finding the table in the internal models due to the ignore of table names with underscores.

DSO-API WFS view

N.A.

N.A.

N.A.

### DIFF
--- a/src/tests/test_dynamic_api/test_views_wfs.py
+++ b/src/tests/test_dynamic_api/test_views_wfs.py
@@ -19,7 +19,7 @@ class TestDatasetWFSView:
     def test_wfs_view(self, api_client, filled_router, afval_dataset, afval_container):
         wfs_url = (
             "/v1/wfs/afvalwegingen/"
-            "?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAMES=containers"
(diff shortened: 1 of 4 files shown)


src/tests/test_dynamic_api/test_views_wfs.py
```diff
@@ -92,3 +92,18 @@ def test_wfs_field_auth(
...
+        it can be useds find the correct table name and data
+        """
+        print("******", afval_adresloopafstand.__dict__)
+        wfs_url = (
+            "/v1/wfs/afvalwegingen/"
+            "?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&TYPENAMES=app:adres_loopafstand"
+            "&OUTPUTFORMAT=application/gml+xml"
+        )
+        response = api_client.get(wfs_url)
+        assert response.status_code == 200
```
ngaranko, line 109: I think checking status is not enough anymore (as error in getting details results in same 200, but with error in xml).
Would be good to check if model is loaded and at least empty list is given back. Ideally create instance of AdresLoopafstand and check if it is returned as WFS.

cvriel: Check! Added extra tests based on presence of output (content)

PR #213, opened 2021-03-05
bug fix AUTH specification on table

cvriel:
In the permissions.py the method `fetch_scopes_for_dataset_table` used the table id  - as specified in the JSON schema - as lookup value for DatasetTable object to get the objects auth attribute. Due to camelCase the object could not be found. 

The object was present in it's snake_case format. As a fix the lookup was added with a snake_case translation. 

In the future a general approach of using snake_case and camelCase in the dso-api logic could be revisited.

permissions.py, fetch_scopes_for_dataset_table method

N.A.

N.A.

N.A.

### DIFF
--- a/src/dso_api/dynamic_api/permissions.py
+++ b/src/dso_api/dynamic_api/permissions.py
@@ -5,6 +5,7 @@ from cachetools.func import ttl_cache
 from django.contrib.auth.models import AnonymousUser, _user_has_perm
 from rest_framework import permissions
 from schematools.contrib.django import models
+from schematools.utils import to_snake_case
 
 from dso_api.dynamic_api.utils import snake_to_camel_case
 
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/permissions.py
```diff
@@ -21,6 +22,13 @@ class TableScopes:
 def fetch_scopes_for_dataset_table(dataset_id: str, table_id: str):
     """ Get the scopes for a dataset and table, based on the Amsterdam schema information """
 
+    # Make sure that names are snake_cased
+    # if used for a lookup in models.DatasetTable.objects
+    # TODO: do snake_case transformations at one centralized place to
+    # to be used in code (as much as possible)
+    dataset_id = to_snake_case(dataset_id)
```
jjmurre, line 29: For now, this is a quick solution. As you already stated, we need to analyze the useage of the snake and camelcasing through the whole application (see issue 5207). The reason we need to do this the snakecasing here, is that the dataset name is snaked_cased before it it stored in the database (see DatasetTable model in schema-tools). Is that the right approach? These and other questions need to be addressed.

jjmurre (approved):
Fixes the problem for now. We need to analyze the casing issue (5207) through and through.

ngaranko (approved):

src/dso_api/dynamic_api/permissions.py
cvriel: I will post this context into the ticket, thnx

PR #215, opened 2021-03-11
cleanup serializer_factory logic, same code flow but more readable

vdboor:
The links_serializer=True check is moved into a `links_serializer_factory()`. Likewise, `_build_serializer_field()` is split into separate methods for relations. This fixes the code complexity issues.
    
The new separate code paths avoids building unnecessary fields for the _links serializer or "body serializer" part that would later be removed in get_fields(). This improves upon the work in #209 / 2dfdf75. The field removal logic is still preserved, as it's used for some situations like detail vs list view.
    
Misc:
* Removed 'depth' testing as that was always true, when testing on other variables early.
* Removed usage of internal _table_schema variables.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -3,11 +3,13 @@ from __future__ import annotations
 import re
 from collections import OrderedDict
 from functools import lru_cache
-from typing import Type
+from typing import Optional, Tuple, Type, Union
 from urllib.parse import quote, urlencode
 
 from django.core.exceptions import ImproperlyConfigured
 from django.db import models
+from django.db.models.fields.related import RelatedField
+from django.db.models.fields.reverse_related import ForeignObjectRel
 from django.utils.functional import cached_property
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
@@ -193,7 +195,7 @@ class DynamicSerializer(DSOModelSerializer):
             field_kwargs["view_name"] = get_view_name(model_class, "detail")
 
         # Only make it a temporal field if model is from a temporal dataset
-        if field_class == HyperlinkedRelatedField and model_class._table_schema.is_temporal:
+        if field_class == HyperlinkedRelatedField and model_class.table_schema().is_temporal:
             field_class = TemporalHyperlinkedRelatedField
         return field_class, field_kwargs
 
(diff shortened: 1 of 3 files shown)


src/dso_api/dynamic_api/serializers.py
```diff
@@ -388,14 +398,19 @@ def get_view_name(model: Type[DynamicModel], suffix: str):
...
 @lru_cache()
-def serializer_factory(
+def serializer_factory(  # noqa: C901
     model: Type[DynamicModel], depth: int, flat=None, links_serializer=False
 ) -> Type[DynamicSerializer]:
     """Generate the DRF serializer class for a specific dataset model.
     It can generate the serializer for both the body as well as for the _links field,
     depending on links_serializer being False or True."""
+    if isinstance(model, str):
+        raise ImproperlyConfigured(f"Model {model} could not be resolved.")
```
ngaranko, line 408: Nice addition, though this check should be probably done before `serializer_factory` is called, as having model name is not enough to find root cause of a problem.

src/dso_api/dynamic_api/serializers.py
```diff
@@ -388,14 +398,19 @@ def get_view_name(model: Type[DynamicModel], suffix: str):
...
 
-    if links_serializer:
```
ngaranko, line 398: Just a brainfart here: maybe it's time to actually split factory into 3 different ones: 
 - `serializer_factory`
 - `flat_serializer_factory`
 - `links_serializer_factory`
Otherwise single factory becomes too complex and prone to be broken by minor changes.

src/dso_api/dynamic_api/serializers.py
vdboor: apparently that was already in the code, but below the `fields = [..]` logic..

src/dso_api/dynamic_api/serializers.py
vdboor: great idea! Makes the code much cleaner too.

src/dso_api/dynamic_api/serializers.py
ngaranko: Yeah, it's an old one and definitely outside of cleanup PR ;)

ngaranko (approved):
LGTM. Nice work!

ngaranko:
:+1:

PR #216, opened 2021-03-11
Optimize querysets, allow prefetch-related on large sets

vdboor:
This performs prefetches in batches, leveraging Django's prefetch_related_objects() while using QuerySet.iterator() for performance.

As extra feature, the previously prefetched objects are reused so the next batch doens't have to refetch identical related objects.

### DIFF
--- a/src/rest_framework_dso/embedding.py
+++ b/src/rest_framework_dso/embedding.py
@@ -6,8 +6,12 @@ the main objects are inspected while they are consumed by the output stream.
 """
 from __future__ import annotations
 
+from collections import defaultdict
+from itertools import islice
 from typing import Callable, Dict, Iterable, Iterator, List, Optional, TypeVar, Union
 
+from django.db import models
+from lru import LRU
 from rest_framework import serializers
 from rest_framework.exceptions import ParseError, PermissionDenied
 
@@ -17,6 +21,8 @@ from rest_framework_dso.serializer_helpers import ReturnGenerator, peek_iterable
 
 EmbeddedFieldDict = Dict[str, AbstractEmbeddedField]
 T = TypeVar("T")
+M = TypeVar("M", bound=models.Model)
+DEFAULT_SQL_CHUNK_SIZE = 2000  # allow unit tests to alter this.
 
 
 def get_expanded_fields(  # noqa: C901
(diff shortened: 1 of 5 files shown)


src/rest_framework_dso/embedding.py
```diff
@@ -77,6 +83,96 @@ def _expand_parse_error(allowed_names, field_name):
...
+        self._persist_prefetch_cache(instances)
+
+    def _persist_prefetch_cache(self, instances):
+        """Store the prefetched data so it can be applied to the next batch"""
+        for instance in instances:
+            for lookup, obj in instance._state.fields_cache.items():
+                if obj is not None:
+                    cache = self._fk_caches[lookup]
+                    if obj.pk not in cache:
+                        cache[obj.pk] = obj
```
lbam, line 148: Is there a reason why you're looking before you leap? This way, the cache is no longer truly LRU. E.g.,

```python
d = LRU(2)
for key in "foo bar baz foo quux foo bar".split():
    if key not in d:
        d[key] = None
```

stores `["bar", "quux"]` in the cache, rather than `["bar", "foo"]`.

vdboor: Thanks, fixed!

src/rest_framework_dso/embedding.py
```diff
@@ -77,6 +83,95 @@ def _expand_parse_error(allowed_names, field_name):
...
+        self._fk_caches = defaultdict(lambda: LRU(self.chunk_size // 2))
+
+    def __iter__(self):
+        # Using iter() ensures the ModelIterable is resumed with the next chunk.
+        qs_iter = iter(self.queryset.iterator(chunk_size=self.sql_chunk_size))
+
+        # Keep fetching chunks
+        while instances := list(islice(qs_iter, self.chunk_size)):
+            # Perform prefetches on this chunk:
+            if self.queryset._prefetch_related_lookups:
```
jjmurre, line 123: Maybe a bit of a risk, to rely on internal Django variables? Otoh, upgrading Django is not automatic, so we will detect it when this changes.

jjmurre (approved):
Nice approach with the prefetch in batches!

src/rest_framework_dso/embedding.py
vdboor: I doubt this will change that fast, and as last resort we could even remove it if this becomes unportable.
The unit test does cover this case specifically.

PR #217, opened 2021-03-11
CSV expand support

vdboor:
The `DSOSerializer` incorporates the embedded-field serializers as regular fields when the output format needs to use inline embedding (e.g. for CSV). To make this performant, the prefetch was optimized in #216 (5e4b9f18f567f3ffddd2a2ccedeb86e6e5ee0515).

The serializer construction is improved to handle sub-resources well, and to avoid expanding their parameters. The many=True is needed to properly initialize inline M2M fields for an output renderer.

All output renderers can declare whether they support "inline" embeds, and whether the support M2M relations.


jjmurre (approved):
Ziet er goed uit, mooi dat we nu ook de embedded field mee kunnen exporteren!

PR #219, opened 2021-03-15
Fix DSOOrderingFilter's snake-casing

lbam:
The get_ordering method assumed it could pass empty identifiers to to_snake_case, which in the latest version refuses to convert those.

### DIFF
--- a/src/rest_framework_dso/filters.py
+++ b/src/rest_framework_dso/filters.py
@@ -642,9 +642,11 @@ class DSOOrderingFilter(OrderingFilter):
         if ordering is None:
             return ordering
 
-        # convert to snake_case, preserving `-` if needed
-        correct_ordering = ["-".join([to_snake_case(y) for y in x.split("-")]) for x in ordering]
-        return correct_ordering
+        # Convert identifiers to snake_case, preserving `-` (descending sort).
+        return [
+            "-" + to_snake_case(part[1:]) if part.startswith("-") else to_snake_case(part)
+            for part in ordering
+        ]
 
     def remove_invalid_fields(self, queryset, fields, view, request):
         """Raise errors for invalid parameters instead of silently dropping them."""


gkoller (approved):

gkoller:
LGTM

src/rest_framework_dso/filters.py
```diff
@@ -642,9 +642,11 @@ def get_ordering(self, request, queryset, view):
         if ordering is None:
             return ordering
 
-        # convert to snake_case, preserving `-` if needed
-        correct_ordering = ["-".join([to_snake_case(y) for y in x.split("-")]) for x in ordering]
-        return correct_ordering
+        # Convert identifiers to snake_case, preserving `-` (descending sort).
+        return [
```
ngaranko, line 646: Part slicing is dangerous. Say we will have a field ('test-data'). This code will break things.

Maybe following will do?:
```
        correct_ordering = ["-".join([to_snake_case(y) for y in x.split("-") if y]) for x in ordering]
````

ngaranko: OR: update `to_snake_case` to ignore empty strings? Probably not such a good idea, but worth discussing

lbam: Are such field names allowed at all? If so, to_snake_case is broken, because it returns test_data, so we have a much bigger problem.

ngaranko: No, they are not. But again, slicing is no good. It won't save us if someboy uses `--name` as sort parameter.

ngaranko: Oh, i see now that my brainfart code is bad idea :)  as it will drop `-` from `-name` filter.

ngaranko (approved):
LGTM.

PR #220, opened 2021-03-15
Include a GeoJSON id in the output, similar to what the WFS server does.

vdboor:
This PR is to solve some differences between the WFS and DSO API output.
It also allows to distinguish between normal and additionally returned objects when `?_embed=true` or `?_embedScope=..` is used.


cvriel (approved):
Lijkt mij prima

PR #223, opened 2021-03-18
AD#5017. Dataset versioning

ngaranko:

Router logic, making sure that non-default datasets are not exposed via API.


DATASET_VERSIONING.md
```diff
@@ -0,0 +1,76 @@
...
+will be treated as not versioned and all tables within dataset will have API endpoint set to 
+
+> `/v1/<dataset.url_prefix>/<dataset.id>/<table.name>/`
+
+Database tables of this dataset will be prefixed with `<dataset.id>_`
+
+
+## Datasets with non-default versions
+
+Any dataset wit `version` defined and `is_default_version` is set to `False` will have API endpoint defined as:
```
gkoller, line 32: Any dataset wit**h** ...

`is_default_version` ~is~ set to ...

DATASET_VERSIONING.md
```diff
@@ -0,0 +1,76 @@
...
+
+ > `/v1/<dataset.url_prefix>/<dataset.id>@<dataset.version>/<table.name>/`
```
gkoller, line 34: Does the API endpoint still start with `/v1`? If so, what does the `v1` stand for in relation the `<dataset.version>`?

DATASET_VERSIONING.md
ngaranko: :+1:

DATASET_VERSIONING.md
ngaranko: Yes, as it is API version, not dataset version. `/v1` is expected to be changed when we change JSON structure drastically.

DATASET_VERSIONING.md
```diff
@@ -0,0 +1,76 @@
...
+
+All tables within default version of Dataset will have `<dataset.id>_` name prefix.
+
+Other Dataset versions will have `<dataset.id>_<dataset.version.major>_<dataset.version.minor>_` table name prefix.
+
+> DRAFT NOTE:
+> 
+>`<dataset.version.major>_<dataset.version.minor>` should be used with caution 
+as underscore can be lost in `to_snake_case` -> `toCamelCase` -> `to_snake_case` translation, 
+however should not come into play as table name is rather isolated.
```
gkoller, line 47: Why not circumvent the whole `to_snake_case` -> `toCamelCase` -> `to_snake_case` problem by using `$` as the separator between major and minor version number? PostgreSQL [allows for it](https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS):

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.

It's ugly to use `$` but anything better than having major and minor version number suddenly being lumped together. (eg `1_3` suddenly becoming `13`)

DATASET_VERSIONING.md
gkoller: Nevermind. It's already clear to me.

DATASET_VERSIONING.md
ngaranko: Dollar sign will do as well, however it might bring more issues in future. 

We do not expect database table name conversions, so underscore should safe to use and less ugly than dollar sign.

DATASET_VERSIONING.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Examples:
+
+Given following Dataset definition:
+
+ - `test` dataset with version `0.1.0` and single table `users` (fields: `id`, `name`) defined
+ - `test` dataset with version `0.1.1` is default and single table `users` (fields: `id`, `name`, `age`) defined
+ - `test` dataset with version `0.2.1` and 2 tables defined: `users` (fields: `id`, `firstName`, `lastName`) and `locations` (fields: `id`, `name`)
```
gkoller, line 55: An extra field to an existing table only increases the patch level number. But an extra table warrants an minor level number increase. Why?

ngaranko: `users` table has changed, I should have highlighted that better.

Will add version changing rules to the top section.

gkoller: What i mean is that logically additions, whether being fields or tables, should be safe -> increase patch number.

All renames, type changes, deletions, are not -> increase minor (and/or major) numbers.

Or am I missing something?

ngaranko: You are absolutely correct. However in example above table ‘users’ has changed structure.

DATASET_VERSIONING.md
```diff
@@ -0,0 +1,76 @@
...
+Any dataset with `version` defined and `is_default_version` set to `False` will have API endpoint defined as:
+
+ > `/v1/<dataset.url_prefix>/<dataset.id>@<dataset.version>/<table.name>/`
+
+
+## Database structure
+
+All tables within default version of Dataset will have `<dataset.id>_` name prefix.
+
+Other Dataset versions will have `<dataset.id>_<dataset.version.major>_<dataset.version.minor>_` table name prefix.
```
lbam, line 41: I think we should call attention to the fact that the patch level is *not* in the table name.

DATASET_VERSIONING.md
lbam: > We do not expect database table name conversions

But if the need does arrive, we've got a big problem. We can plan for that, though, by putting the version last: `test_users_0_2` instead of `test_0_2_users` (or even `test_users_v0_2`?). That way, it'll be easier to parse the version number off the table name before attempting any conversion.

ngaranko: Using version suffix might be better approach. 
Though it has disadvantage: it will be harder to search for tables that belong to same version and that might be bigger issue with cross-dataset relation tables, where version of both datasets should be taken into account.

DATASET_VERSIONING.md
lbam: If any field gets deleted, renamed or has its type changed, that's a breaking change for consumers and SemVer calls for a major version increment.

Adding a field should be a minor increment, I think.

In this example, the major version never actually changes...

ngaranko: Should it? This is not versioning example, this is database & api structure behavior. 

Document on schema versioning is going to be added to amsterdam-schema repo and outside of this document.

lbam: Of course it's a versioning example, it's an example that does versioning :)

DATASET_VERSIONING.md
lbam: You mean because `users_.*_1_0` matches `users_id_1_1__dogs_id_1_0`?

DATASET_VERSIONING.md
ngaranko: Unfortunately we do not know how versioning will be working just yet. 

Major version can be reserved to dataset owner to decide if bump is need or maybe major bump comes with change of dataset source.

Versioning is up for discussion. Goal of this document is to outline what happens with database if versions change.

lbam: Alright. I thought we did Semver, but I just saw that you wrote "formatted according to" :)

DATASET_VERSIONING.md
```diff
@@ -0,0 +1,84 @@
+# Dataset versioning (DRAFT)
```
vdboor, line 1: Can we move this into the `dev-docs`?

src/dso_api/dynamic_api/routers.py
```diff
@@ -264,6 +265,13 @@ def _prune_app_registry(self, old_dynamic_apps: set):
             del apps.all_models[removed_app]
 
     def filter_datasets(self, queryset):
+        """Filter datasets:
+        - remove Non-default datasets
+        - include only datasets defined in DATASETS_LIST (if settings.DATASETS_LIST is defined)
+        - exclude any datasets in DATASETS_EXCLUDE list (if settings.DATASETS_EXCLUDE is defined)
+        """
+        queryset = queryset.filter(Q(version=None) | Q(is_default_version=True))
```
vdboor, line 273: Can we add a manager/queryset method to schematools: `queryset.default_version()`?

src/tests/test_dynamic_api/test_router.py
```diff
@@ -92,3 +92,14 @@ def test_router_excludes_datasets_combined_with_list(
         assert reverse("dynamic_api:bommen-bommen-list")
     with pytest.raises(NoReverseMatch):
         assert reverse("dynamic_api:meldingen-statistieken-list")
+
+
+@pytest.mark.django_db
+def test_router_excludes_non_default_dataset_versions(settings, router, bommen_v2_dataset):
+    router.reload()
```
vdboor, line 99: after my faster tests merge:

```suggestion
def test_router_excludes_non_default_dataset_versions(settings, bommen_v2_dataset, filled_router):
```

src/dso_api/dynamic_api/routers.py
ngaranko: This is only place where default version filtering is applied. And queryset/manager is a bit of overkill.

src/tests/test_dynamic_api/test_router.py
ngaranko: But it is not merged yet :)

ngaranko: En zo heb ik het gedaan:
```
def test_router_excludes_non_default_dataset_versions(settings, bommen_v2_dataset, filled_router):
```

>>

```
E                   RuntimeError: The 'filled_router' or 'dynamic_models' fixture was requested, but no dataset fixtures were defined for this test
```
:(

ngaranko: Added comment on why `filled_router` can't be used in this case.

DATASET_VERSIONING.md
ngaranko: Moved, thanks for notice.

src/tests/files/bommen.json
```diff
@@ -3,6 +3,7 @@
   "type": "dataset",
   "title": "",
   "version": "0.0.1",
+  "version": "0.0.1",
```
vdboor, line 6: duplicate key!

src/tests/test_dynamic_api/test_views_wfs.py
```diff
@@ -112,3 +113,47 @@ def test_wfs_feature_name(self, api_client, afval_dataset, afval_adresloopafstan
...
+        api_client,
+        router,
+        bommen_dataset,
+    ):
+        """Prove that if feature name contains non-letters like underscore,
+        it can be useds find the correct table name and data
+        """
+        router.reload()
+        # manually creating tables, as we do not want to use `filled_router` here.
+        create_tables(bommen_dataset)
```
vdboor, line 128: use `filled_router` here?

src/tests/test_dynamic_api/test_views_wfs.py
```diff
@@ -112,3 +113,47 @@ def test_wfs_feature_name(self, api_client, afval_dataset, afval_adresloopafstan
...
+        self,
+        api_client,
+        router,
+        bommen_dataset,
+    ):
+        """Prove that if feature name contains non-letters like underscore,
+        it can be useds find the correct table name and data
+        """
+        router.reload()
+        # manually creating tables, as we do not want to use `filled_router` here.
```
vdboor, line 127: why? :-)  (perhaps useful to add in the comment)

ngaranko: I'll  copy paste comment from another test. Good one.

src/tests/files/bommen.json
ngaranko: Eagle eye! Fixing.

PR #224, opened 2021-03-18
guido/ab5505 k8s tracing logging monitoring

gkoller:
Tracing and logging configuration for when the DSO API will be run in an Azure environment

Be sure to read the README.md for how to configure this. In short it requires the following env vars to be set (replace `...` with meaningful values.

```
CLOUD_ENV=azure
AZURE_APPI_INSTRUMENTATION_KEY=....
AZURE_APPI_INGESTION_ENDPOINT=...
```

### DIFF
--- a/src/dso_api/settings.py
+++ b/src/dso_api/settings.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import sys
-from enum import Enum
 from pathlib import Path
 from typing import Optional
 
@@ -14,16 +13,6 @@ from opencensus.trace import config_integration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import LoggingIntegration
 
-
-class CloudEnv(Enum):
-    CLOUDVPS_DOCKER = "cloudvps_docker"
-    AZURE_ACI = "azure_aci"
-    AZURE_AKS = "azure_aks"
-
-    def is_azure(self):
-        return self in (CloudEnv.AZURE_ACI, CloudEnv.AZURE_AKS)
-
-
 env = environ.Env()
 
 # -- Environment
@@ -31,7 +20,7 @@ env = environ.Env()
 BASE_DIR = Path(__file__).parents[1]
 DEBUG = env.bool("DJANGO_DEBUG", True)
 
-CLOUD_ENV = CloudEnv(env.str("CLOUD_ENV", "cloudvps_docker"))
+CLOUD_ENV = env.str("CLOUD_ENV", "unspecified")
 DJANGO_LOG_LEVEL = env.str("DJANGO_LOG_LEVEL", "INFO")
 DSO_API_LOG_LEVEL = env.str("DSO_API_LOG_LEVEL", "INFO")
 DSO_API_AUDIT_LOG_LEVEL = env.str("DSO_API_AUDIT_LOG_LEVEL", "INFO")
@@ -210,7 +199,7 @@ LOGGING = {
     },
 }
 
-if CLOUD_ENV.is_azure():
+if CLOUD_ENV.lower().startswith("azure"):
     if AZURE_APPI_INSTRUMENTATION_KEY is None:
         raise ImproperlyConfigured(
             "Please specify the 'AZURE_APPI_INSTRUMENTATION_KEY' environment variable."
--- a/src/dso_api/settings.py
+++ b/src/dso_api/settings.py
@@ -91,31 +91,6 @@ MIDDLEWARE = [
     "dso_api.dynamic_api.middleware.TemporalDatasetMiddleware",
     "dso_api.dynamic_api.middleware.RequestAuditLoggingMiddleware",
 ]
-connection_string: Optional[str] = None
-if CLOUD_ENV.is_azure():
-    if AZURE_APPI_INSTRUMENTATION_KEY is None:
-        raise ImproperlyConfigured(
-            "Please specify the 'AZURE_APPI_INSTRUMENTATION_KEY' environment variable."
-        )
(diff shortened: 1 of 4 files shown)


src/dso_api/settings.py
```diff
@@ -68,6 +89,31 @@
     "dso_api.dynamic_api.middleware.TemporalDatasetMiddleware",
     "dso_api.dynamic_api.middleware.RequestAuditLoggingMiddleware",
 ]
+connection_string: Optional[str] = None
```
ngaranko, line 92: Maybe it would be better to move whole block down below `LOGGING` definition, then we would not need 2 identical `if CLOUD_ENV.is_azure():` statements and probably can get rid of temporary variable `connection_string`.

src/dso_api/settings.py
```diff
@@ -25,7 +43,10 @@
 PROFILES_URL = env.str("PROFILES_URL", "https://schemas.data.amsterdam.nl/profiles/")
 SCHEMA_DEFS_URL = env.str("SCHEMA_DEFS_URL", "https://schemas.data.amsterdam.nl/schema")
 
+# -- Azure specific settings
 
+AZURE_INSTRUMENTATION_KEY: Optional[str] = env.str("AZURE_INSTRUMENTATION_KEY", None)
```
ngaranko, line 48: Let's be more specific here and add Azure service name into variable. Otherwise soon it will become confusing which azure variable does what.

src/dso_api/settings.py
```diff
@@ -176,7 +221,7 @@
     },
     "root": {"level": "INFO", "handlers": ["console"]},
     "loggers": {
-        # "django.db.backends": {"level": "DEBUG", "handlers": ["console"]},
+        "django": {"handlers": ["console"], "level": DJANGO_LOG_LEVEL, "propagate": False},
         "dso_api": {"handlers": ["console"], "level": "DEBUG", "propagate": False},
```
ngaranko, line 225: Unrelated to current PR (not totally, but does not have to be fixed now): do we really need `DEBUG` logging on production?

src/dso_api/settings.py
gkoller: Sure no problem. Though the downside of that approach is that the middleware configuration will then happen at two separate locations in `settings.py`: one where `MIDDLEWARE` is defined, and the other one some time after the logging configuration

ngaranko: That's okay, because additional MIDDLEWARE is added conditionally.

gkoller: done

src/dso_api/settings.py
gkoller: Changed default level to `INFO` and made it configurable by means of an environment variable.

src/dso_api/settings.py
gkoller: Done

ngaranko: Thanks a lot! Though I meant Azure service name, as you know we have quite few and they all are looking similar.

vdboor (approved):

src/dso_api/settings.py
```diff
@@ -176,16 +200,61 @@
...
         "dso_api.audit": {
             "handlers": ["structured_console"],
-            "level": "DEBUG",
+            "level": DSO_API_AUDIT_LOG_LEVEL,
             "propagate": False,
         },
     },
 }
 
+if CLOUD_ENV.is_azure():
```
vdboor, line 213: `if CLOUD_ENV.startswith("azure")` had ik ook goed genoeg gevonden :-p laat je het wat meer open en zijn de environments niet beperkt tot de enum.

src/dso_api/settings.py
```diff
@@ -1,21 +1,41 @@
...
 
+
+class CloudEnv(Enum):
+    CLOUDVPS_DOCKER = "cloudvps_docker"
+    AZURE_ACI = "azure_aci"
+    AZURE_AKS = "azure_aks"
+
+    def is_azure(self):
+        return self in (CloudEnv.AZURE_ACI, CloudEnv.AZURE_AKS)
+
```
vdboor, line 25: Erg cool, tegelijk een beetje information leakage en maakt het minder herbruikbaar voor anderen. Misschien gewoon een string laten?

ngaranko (approved):

src/dso_api/settings.py
gkoller: This has the advantage that the strings are verified by virtue of https://github.com/Amsterdam/dso-api/pull/224/files#diff-f1b9d231aef12e16d24d3da417f90c4471fd4d9a213446ed5579bef76d50a992R34

Meaning if the user provides `CLOUD_ENV=fubar` the call `CloudEnv(env.str("CLOUD_ENV", "cloudvps_docker"))` will fail.

gkoller: BTW not sure what you mean with information leakage. What are we leaking? Possible values? But we leak that anyway when we later on test for specific values, regardless of whether they are string or enums.

src/dso_api/settings.py
gkoller: In that case I'd rather adjust the definition of `CloudEnv.is_azure()` to:

```python
    def is_azure(self):
        azure = (v for v in CloudEnv.__members__.values() if v.value.startswith("azure"))
        return self in azure
```

gkoller: Limiting the possible environment variables to the enum was done on purpose; to prevent non-sensical values. I don't want devops engineers to set `CLOUD_ENV=aks` and expecting it to work. I want to have the application fail early so that it can be troubleshooted immediately

gkoller: Discussed it offline. I now agree with @vdboor. Ditched the `Enum`

PR #252, opened 2021-04-22
Integration test for HC-BRK

lbam:
This adds a small integration test for HC-BRK (Kadaster), using the example output produced by the Kadaster API. It can be run using pytest.

I wasn't sure whether to put this in `src` or at toplevel. Thoughts?


ngaranko (approved):
LGTM

PR #256, opened 2021-05-03
Test multiple required scopes in MVT auth test

lbam:

Authorization tests. The MVT test now requires two scopes to access the geometry field.

### DIFF
--- a/src/tests/test_dynamic_api/test_views_mvt.py
+++ b/src/tests/test_dynamic_api/test_views_mvt.py
@@ -72,15 +72,8 @@ def test_mvt_forbidden(api_client, geometry_auth_thing, fetch_auth_token, filled
 def test_mvt_model_auth(api_client, geometry_auth_thing, fetch_auth_token, filled_router):
     """Prove that unauthorized fields are excluded from vector tiles"""
 
-    # Get authorization for the geometry field, but not the meta field.
-    token = fetch_auth_token(["TEST/GEO"])
-
     url = "/v1/mvt/geometry_auth/things/1/0/0.pbf"
-    response = api_client.get(url, HTTP_AUTHORIZATION=f"Bearer {token}")
-    assert response.status_code == 200
-    assert response["Content-Type"] == CONTENT_TYPE
-
-    assert mapbox_vector_tile.decode(response.content) == {
+    content = {
         "default": {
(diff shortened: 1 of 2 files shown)


src/tests/test_dynamic_api/test_views_mvt.py
```diff
@@ -72,10 +72,15 @@ def test_mvt_forbidden(api_client, geometry_auth_thing, fetch_auth_token, filled
...
-    token = fetch_auth_token(["TEST/GEO"])
-
     url = "/v1/mvt/geometry_auth/things/1/0/0.pbf"
+
+    # The geometry field requires two scopes. Giving it only one should give 403.
+    token = fetch_auth_token(["TEST/GEO1"])
+    response = api_client.get(url, HTTP_AUTHORIZATION=f"Bearer {token}")
+    assert response.status_code == 403
+
+    # Even with both scopes, the meta field should still not be accessible.
```
ngaranko, line 82: `not be accessible.` but gives `200` :)

lbam: The request should still succeed, but with the field missing form the response.

gkoller: I'm a big fan of using the `HTTPStatus` enum. Might I suggest:

```python
assert response.status_code == HTTPStatus.FORBIDDEN
```

lbam: We don't use that anywhere. I'd rather keep this consistent with the rest of the tests for now. We can migrate the whole codebase to HTTPStatus in a later PR.

ngaranko: Please update comment to represent expected: `The request should still succeed, but with the field missing form the response.`
Otherwise it is confusing.

ngaranko: PS: maybe adding

 `assert "metadata" not in mapbox_vector_tile.decode(response.content)["default"]["features"].keys()`

 would bring much more clarity to what this test is about.

lbam:
I rewrote the test so that it checks for full access first, then explicitly drops a scope and a field. Please review again.

ngaranko (approved):

PR #257, opened 2021-05-03
Run WFS auth tests on custom schemas

lbam:

WFS authorization tests now use schemas instead of a model that is modified on the fly. Follow-up to #255.

### DIFF
--- a/src/tests/test_dynamic_api/test_views_wfs.py
+++ b/src/tests/test_dynamic_api/test_views_wfs.py
@@ -1,5 +1,6 @@
+from typing import List
+
 import pytest
-from schematools.contrib.django import models
 from schematools.contrib.django.db import create_tables
 
 from dso_api.dynamic_api.permissions import fetch_scopes_for_dataset_table, fetch_scopes_for_model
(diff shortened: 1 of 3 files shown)


src/tests/test_dynamic_api/test_views_wfs.py
```diff
@@ -180,3 +124,72 @@ def test_wfs_non_default_dataset_not_exposed(
...
+        "boundedBy": None,
+        "geometry": None,
+        "id": "1",
+        "metadata": "secret",
+    }
+
+
+@pytest.mark.django_db
+def test_wfs_field_auth(api_client, geometry_auth_thing, fetch_auth_token, filled_router):
+    def get_response(*scopes):
```
ngaranko, line 161: It is good practice to avoid any complexity in tests, and better to split test into 3 different tests instead of reusing `get_response` function inside.

Splitting test into 3 different functions would make tests more readable and also will help isolating issues in code a bit easier. 
Single  use case tests in combination clear names for tests save ton of time when debugging broken code.

jjmurre: Maybe split up the different testcases using `pytest.parametrize`.

src/tests/test_dynamic_api/test_views_wfs.py
```diff
@@ -180,3 +124,72 @@ def test_wfs_non_default_dataset_not_exposed(
         )
         response = api_client.get(wfs_url)
         assert response.status_code == 404
+
+
+@pytest.mark.django_db
+def test_wfs_model_auth(api_client, geometry_authdataset_thing, fetch_auth_token, filled_router):
```
vdboor, line 130: I'd prefer if the tests remain grouped. When you revisit a test file it's nearly impossible to see where to insert the next test. Grouping in classes helps for this.

src/tests/test_dynamic_api/test_views_wfs.py
vdboor: This is indeed a good case for `pytest.mark.parametrize`.

Or
* move the `get_response()` outside the function
* have 3 separate test calls. 
* all grouped in a single class

lbam:
I tried to address your comments. Please see updated code.

ngaranko (approved):

PR #259, opened 2021-05-04
Remove comma-separated "auth" list support

lbam:
"auth" fields on datasets, tables and fields could previously be comma-separated lists stored in JSON strings (`"FOO/BAR,FOO/BAZ"`). This feature was never used and is now removed.

### DIFF
--- a/src/dso_api/dynamic_api/permissions.py
+++ b/src/dso_api/dynamic_api/permissions.py
@@ -14,8 +14,10 @@ from schematools.utils import to_snake_case, toCamelCase
 class TableScopes:
     """OAuth scopes for tables and fields."""
 
-    table: Set[str] = field(default_factory=set)
-    fields: Dict[str, Set[str]] = field(default_factory=dict)
+    # Table and dataset scopes.
+    table: Set[str] = field(default=None)
+    # Scope per field.
+    fields: Dict[str, str] = field(default_factory=dict)
 
 
 @ttl_cache(maxsize=None, ttl=60 * 60)
@@ -29,19 +31,14 @@ def fetch_scopes_for_dataset_table(dataset_id: str, table_id: str):
     dataset_id = to_snake_case(dataset_id)
     table_id = to_snake_case(table_id)
 
-    def _fetch_scopes(obj):
-        if obj.auth:
-            return set(obj.auth.split(","))
-        return set()
-
     try:
         table = models.DatasetTable.objects.get(name=table_id, dataset__name=dataset_id)
     except models.DatasetTable.DoesNotExist:
         return TableScopes()
 
     return TableScopes(
-        table=_fetch_scopes(table) | _fetch_scopes(table.dataset),
-        fields={field.name: _fetch_scopes(field) for field in table.fields.all()},
+        table=frozenset(level.auth for level in [table, table.dataset] if level.auth is not None),
+        fields={field.name: field.auth for field in table.fields.all() if field.auth is not None},
     )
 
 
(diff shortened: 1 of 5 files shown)


jjmurre (approved):
Ziet er goed uit. Wat simpeler zo, altijd goed :-)

lbam:
Ik heb nog een docstring preciezer gemaakt en een type annotation toegevoegd aan `get_unauthorized_fields`.

src/dso_api/dynamic_api/permissions.py
```diff
@@ -58,19 +55,28 @@ def fetch_scopes_for_model(model: Type[Model]) -> TableScopes:
...
     model_scopes = fetch_scopes_for_model(model)
     unauthorized_fields = set()
-    # is_authorized_for is added by authorization_django middleware
-    if hasattr(request, "is_authorized_for"):
-        for model_field in model._meta.get_fields():
-            scopes = model_scopes.fields.get(model_field.name)
-            scopes = scopes.union(model_scopes.table) if scopes else model_scopes.table
+
+    for model_field in model._meta.get_fields():
+        name = model_field.name
```
ngaranko, line 69: Is this temporary variable really needed? It's used in 2 places.

src/dso_api/dynamic_api/permissions.py
```diff
@@ -58,19 +55,28 @@ def fetch_scopes_for_model(model: Type[Model]) -> TableScopes:
...
+
+        required = set()
+        if (field_scope := model_scopes.fields.get(name)) is not None:
+            required.add(field_scope)
+        required |= model_scopes.table
```
ngaranko, line 74: This does exactly same as following did before, is there real reason for rewrite?

```
scopes = model_scopes.fields.get(model_field.name)
scopes = scopes.union(model_scopes.table) if scopes else model_scopes.table
```

jjmurre: Or `required = model_scopes.fields.get(model_field.name, set()) | model_scopes.table`

lbam: The change is necessary because `model_scopes.fields[name]` is now a `str` instead of the previous `set`. I couldn't come up with an elegant one-liner to create a singleton set for a `str` and an empty set for `None`.

src/dso_api/dynamic_api/permissions.py
lbam: It makes the following code a bit less dense.

src/dso_api/dynamic_api/permissions.py
jjmurre: Aha, I see your point.

jjmurre: You could do this: `required = set(model_scopes.fields.get(model_field.name)) - {None} | model_scopes.table`, but I doubt if that's more readable :-)

lbam: I could write

```py
required = set(s for s in [model_scopes.fields.get(model_field.name)] if s is not None) | model_scopes.table
```

but I'm not sure if that's any better. Maybe

```py
required = set([field_scope]) if (field_scope := model_scopes.fields.get(name)) is not None else set()
required |= model_scopes.table
```

? That's hardly less complicated than the current code.

ngaranko: Following is much more readable, also makes it clear that field scope is singular.
```
 required = model_scopes.table
 field_scope = model_scopes.fields.get(model_field.name)
 if field_scope:
     required.union(set(field_scope))
```


It's nice that `TableScopes` was changed to present fields as dict with strings as values, but then why do we still have table scopes as set, isn't it more confusing now?

src/dso_api/dynamic_api/permissions.py
ngaranko: Can it be named clearly? `field_name` is better, but then again. It's just few characters saved for 2 occasions.

lbam: I've removed it :)

src/dso_api/dynamic_api/permissions.py
lbam: I'll separate the assignment out (the new Python syntax is strange anyway).

The rest of your suggestion is pretty much what I started with. It doesn't work because (1) `set(field_scope)` constructs a set of the characters of `field_scope` (very hard bug to catch) and (2) `union` doesn't modify the left-hand object. `set.add` avoids both problems and is explicit.

`TableScopes.table` is a set because it stores the table scope and/or the dataset scope. I could separate out the dataset scope, but this was already the case.

ngaranko: Oh, sorry for brainfart code :) 

replace: `required.union(set(field_scope))` with `required = required.union([field_scope])`

lbam: Oh I didn't want to start with `required = model_scopes.table` because that makes it easy to accidentally modify the `TableScopes.table` set in-place. Maybe we should make that a `frozenset`? It's not a big deal now, but it's better to be safe than sorry...

lbam: I rewrote it following your suggestion **and** made the set a frozenset.

ngaranko (approved):
Thanks for your patience! :)

lbam:
Merged with one change: the docstring for has_permission still mentioned the comma-separated list feature.

PR #260, opened 2021-05-06
Expand auth docs with scope semantics

lbam:


ngaranko (approved):

dev-docs/source/auth.rst
```diff
@@ -18,11 +19,26 @@ The schema definitions can add an ``auth`` field on various levels:
 * A single table
 * A single field
 
-This defines which fields, tables or datasets are not accessible
-unless a particular authorization scope is provided.
-If the required scope is missing, the fields are omitted from the response.
-
-By default, the REST serializers only return the public fields.
+At every level, the ``auth`` field contains a string that is called a *scope*.
```
jjmurre, line 22: I think the auth field containing a list of scopes is still a possibility (we only removed the komma-separated option). Or, am I wrong? In the meta-schema definition, a list of scopes is possible: https://github.com/Amsterdam/amsterdam-schema/blob/31446ce22b275340a7e99ef3bab988f1db68c24a/schema%40v1.1.1/schema.json#L101

lbam: As far as I can tell, that functionality is missing/broken in dso-api, or only implemented for profiles.

PR #263, opened 2021-05-11
Fix dev-docs about authorization, again

lbam:
The dev-docs incorrectly describe the scope checking as performing an AND of scopes at different levels. That turns out to be false. Only the highest-level scopes are checked.

Also, the dev-docs were kind of fuzzy about the exact workings of profiles. This explains them in some more detail wrt. requests, scopes and schemas. The "statistician" example now explains exactly how schemas fall short (lack of an OR operator, really).

In the example, I've replaced "encoded" by "read" because "encoded" isn't actually implemented in dso-api.

### DIFF
--- a/dev-docs/source/auth.rst
+++ b/dev-docs/source/auth.rst
@@ -45,8 +45,15 @@ Profiles
 
 The ``auth`` fields define the basic rules for authentication.
 Unfortunately, this is a "all or nothing" approach that isn't sufficient in complex cases.
-To overcome this problem, "authentication profiles" were introduced.
-The file format provides various options, such as:
+*Profiles* provide a more fine-grained approach to authorization.
+
+Profiles are stored in the DSO-API database.
+They have a name, a set of scopes, and rules that grant additional permissions
+to requests that carry those scopes in their JWT.
+When a request comes in, all profiles are checked against the request's scopes
+and the matching profiles are applied.
+
+Here's an example profile in JSON:
 
 .. code-block:: json
 
@@ -59,7 +66,7 @@ The file format provides various options, such as:
                     "ingeschrevenpersonen": {
                         "permissions": "read",
                         "fields": {
-                            "bsn": "encoded"
+                            "bsn": "read"
                         },
                         "mandatoryFilterSets": [
                             ["bsn", "lastname"],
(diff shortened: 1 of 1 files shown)


dev-docs/source/auth.rst
```diff
@@ -23,13 +23,14 @@ At every level, the ``auth`` field contains a string that is called a *scope*.
...
 to meet an ``auth`` field's requirements.
-The absence of an ``auth`` field marks a publicly available resource.
+The absence of a scope marks a publicly available resource.
 
-The presence of ``auth`` fields on multiple levels
-means that all listed scopes must be carried in the JWT.
-So, if the dataset has ``"auth": "FOO/BAR"`` and a field has ``"auth": "FOO/BAZ``,
-then access to the field is only available
-with a token that carries the ``FOO/BAR`` and ``FOO/BAZ`` scopes.
+Scopes at higher levels override scopes at lower levels.
```
ngaranko, line 28: *higher levels of schema.
Otherwise it may be confusing people that scopes also have hierarchy, which they do not really.

dev-docs/source/auth.rst
```diff
@@ -23,13 +23,14 @@ At every level, the ``auth`` field contains a string that is called a *scope*.
 A request's JWT must contain this scope in its ``scopes`` claim
 (see the ``authorization_django`` source code for details)
 to meet an ``auth`` field's requirements.
-The absence of an ``auth`` field marks a publicly available resource.
+The absence of a scope marks a publicly available resource.
```
ngaranko, line 26: This was correct, isn't it?

lbam: I thought it wasn't clear enough, but my edit actually made it less clear. Let me try again.

dev-docs/source/auth.rst
lbam: Good point.

dev-docs/source/auth.rst
lbam: Restored to the old text.

dev-docs/source/auth.rst
```diff
@@ -71,18 +78,28 @@ The file format provides various options, such as:
...
 may only access data of someone when they can provide their last name and postal code.
-
-Another case for profiles would be a statistician, that may only read the age and neighbourhood
+Or a statistician might be allowed to read age and neighbourhood
 fields to aggregate data, without ever having access to identifiable data.
-Encoding such feature into the schema file amongst other rules would produce
-very complex ``auth`` fields, while the profile file gives a
-fitting definition of such role and permissions.
+Encoding such rules in the schema file using a custom scope
+would require every other request that accesses the dataset to also have this scope.
```
jjmurre, line 103: May add some explanation about what `"permission": "read"` exactly means compared to `"field": {"bsn": "read"}`.

lbam: I honestly don't know what it means.

dev-docs/source/auth.rst
```diff
@@ -25,11 +25,12 @@ A request's JWT must contain this scope in its ``scopes`` claim
...
-then access to the field is only available
-with a token that carries the ``FOO/BAR`` and ``FOO/BAZ`` scopes.
+Scopes at higher levels of the schema override scopes at lower levels.
+That is, a dataset scope overrides a table scope
+and a table scope overrides field scopes.
+So, if the dataset has ``"auth": "FOO/BAR"``,
+then this scope is required for any access to the dataset
+and any lower-level scopes are ignored.
 
 The effect of an ``auth`` on a dataset or table is that attempts to access it
```
jjmurre, line 35: Maybe add: To access it "without the proper scope"  results in ...

dev-docs/source/auth.rst
ngaranko: Permission "read" means user can see original value of the field in database. (e.g. default for all). It is used to give access to individual fields using Profile.

jjmurre (approved):
Clear explanation!  One other small thing: line 9 of auth.rst mentions "These scopes", but that does not refer to scopes mentioned before in the text.

ngaranko (approved):

lbam:
Thanks!

PR #265, opened 2021-05-17
Guido/ab6678 split audit logging

gkoller:
This PR changes:
- all logging formats to be JSON 
- tags the audit logs with the key-value pair `"audit": true`. 

This should allow filtering of audit specific logs in whatever environment/system we sent our logs to.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -172,10 +172,13 @@ opencensus-ext-azure==1.0.7
     # via -r requirements.in
 opencensus-ext-django==0.7.4
     # via -r requirements.in
+opencensus-ext-logging==0.1.0
+    # via -r requirements.in
 opencensus==0.7.12
     # via
     #   opencensus-ext-azure
     #   opencensus-ext-django
+    #   opencensus-ext-logging
 ordered-set==4.0.2
     # via deepdiff
 orjson==3.5.1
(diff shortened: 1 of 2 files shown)


cvriel (approved):

jjmurre (approved):
I do not know the ins and outs of Opencensus, but the configuration code look good to me.

src/requirements.txt
```diff
@@ -172,10 +172,13 @@ opencensus-ext-azure==1.0.7
     # via -r requirements.in
 opencensus-ext-django==0.7.4
     # via -r requirements.in
+opencensus-ext-logging==0.1.0
+    # via -r requirements.in
```
ngaranko, line 176: I see no changes to `requirements.in` forgot to check in file?

ngaranko (approved):

src/requirements.txt
gkoller: No, it's the other way around ;-). I committed `requirements.in` in commit 9e8d958c91e933d31fd8206e84cb2d1f76e35f5f (March 18th) but forgot to regenerate `requirements.txt`. In this PR I'm fixing that omission.

ngaranko: ;(

PR #266, opened 2021-05-17
Support nested expands + optimize queries

vdboor:
* `?_expand=true` will give 2 levels of expansion now.
* support `?_expandScope=field.subfield`
* the whole way expanded fields are resolved has changed to allow nesting.
* optimized data retrieval to make sure the nesting won't cause performance issues


src/dso_api/dynamic_api/remote/serializers.py
```diff
@@ -43,7 +43,7 @@ class RemoteListSerializer(DSOListSerializer):
 
     @property
     def expanded_fields(self):
```
jjmurre, line 45: Maybe add type annotation. Makes it easier to read the code.

src/dso_api/dynamic_api/serializers.py
```diff
@@ -133,6 +141,22 @@ def to_representation(self, value: models.Manager):
         return {"count": value.count(), "href": f"{url}{query_string}"}
 
 
+class DynamicListSeriallizer(DSOModelListSerializer):
```
jjmurre, line 143: Typo `Seriallizer` -> `Serializer`.

src/rest_framework_dso/embedding.py
```diff
@@ -7,89 +7,212 @@
...
-from rest_framework.exceptions import ParseError, PermissionDenied
+from rest_framework.exceptions import ParseError
 
-from dso_api.dynamic_api.permissions import fetch_scopes_for_model
-from rest_framework_dso.fields import AbstractEmbeddedField, EmbeddedField
+from rest_framework_dso.fields import AbstractEmbeddedField
 from rest_framework_dso.serializer_helpers import ReturnGenerator, peek_iterable
 
-EmbeddedFieldDict = Dict[str, AbstractEmbeddedField]
+DictOfDicts = Dict[str, Dict[str, dict]]
```
jjmurre, line 23: `dict` -> `Dict` (in the embedded dict).

src/rest_framework_dso/embedding.py
```diff
@@ -7,89 +7,212 @@
...
+    if not expand_scope:
+        return {}
+    elif expand_scope is True:
+        return get_all_embedded_field_names(serializer_class, allow_m2m=allow_m2m)
+    else:
+        return group_dotted_names(expand_scope)
+
+
+@dataclass
+class EmbeddedFieldMatch:
```
jjmurre, line 65: Nicely packaged into a dataclass! Makes it very readable indeed.

src/rest_framework_dso/fields.py
```diff
@@ -95,17 +96,16 @@ def is_array(self) -> bool:
 
     @cached_property
     def attname(self):
```
jjmurre, line 98: Maybe add annotation for return type. Make it extra clear that it is heterogeneous.

src/rest_framework_dso/embedding.py
```diff
@@ -224,6 +224,39 @@ def _expand_parse_error(allowed_names, field_name, prefix=""):
         )
 
 
+def get_serializer_lookups(serializer: serializers.BaseSerializer, prefix="") -> List[str]:
```
jjmurre, line 227: Nice! Lot of thinking behind this I guess.

src/rest_framework_dso/embedding.py
```diff
@@ -437,13 +437,27 @@ def inspect_instance(self, instance):
 
     def get_objects(self):
```
jjmurre, line 438: Maybe a return type annotation.

src/rest_framework_dso/embedding.py
vdboor: tx :-)

src/dso_api/dynamic_api/remote/serializers.py
vdboor: tx! :-)

src/rest_framework_dso/embedding.py
vdboor: :-)

src/dso_api/dynamic_api/serializers.py
vdboor: sharp! Fixed!

src/rest_framework_dso/embedding.py
vdboor: It can't use recursion, so I think it's better to leave it like this. `Dict[str, Dict[str, Dict[str, ...]]]` would not be readable.

PR #274, opened 2021-05-26
Fix _links section, hide useless title and null links

vdboor:
* Hide title when the model doesn't have a display field
* Hide null values and empty M2M lists.


jjmurre (approved):
LGTM (nu komen we in de categorie pixelpoetsen :-)

ngaranko (approved):
LGTM

PR #276, opened 2021-05-31
Factor remote endpoint clients out of remote viewsets

lbam:
This is a preliminary PR for HC-BRK authorization. The main functional change is that the HC-BRK endpoint no longer forwards its authorization token to the remote. Doing our own authentication is the next step.

### DIFF
--- a/src/dso_api/dynamic_api/remote/views.py
+++ b/src/dso_api/dynamic_api/remote/views.py
@@ -1,56 +1,21 @@
 """The remote views retrieve data from other REST API endpoints.
 Currently it mainly performs authorization, data retrieval, and schema validation.
-The endpoint is queried with raw urllib3 logic, to avoid the overhead of the requests library.
 """
 import logging
-from typing import Type, Union
-from urllib.parse import urlparse
+from typing import Type
 
-import certifi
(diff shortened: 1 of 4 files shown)


src/dso_api/dynamic_api/remote/views.py
```diff
@@ -141,184 +94,21 @@ def validate(self, serializer, raw_data):
...
 ) -> Type[RemoteViewSet]:
     """Construct the viewset class that handles the remote serializer."""
+
     return type(
         f"{serializer_class.__name__}Viewset",
         (RemoteViewSet,),
         {
             "__doc__": "Forwarding proxy serializer",
-            "endpoint_url": endpoint_url,
+            "client": clients.make_client(endpoint_url, table_schema),
```
vdboor, line 108: I'd use `client_class` instead, and call `make_client` inside the view?

vdboor (approved):

src/dso_api/dynamic_api/remote/views.py
lbam: As long as clients are stateless, we don't need separate instances.

PR #280, opened 2021-06-02
Separate MVT+WFS permissions checks from DRF checks

lbam:
MVT and WFS used an overloaded `has_permission` method to do auth checks. They now call a custom method so the code paths are more clearly separated.


jjmurre (approved):
LGTM. Maybe add a test for it?

lbam:
Please check again, I just pushed a further refactor.

All paths except the else branch in has_permission are tested. I'm not sure what should trigger that path.

jjmurre:
The else branch seems a bit hypothetical. But the `has_permission_for_models` that is being called is tested. So, for now that is good enough I would say.

PR #284, opened 2021-06-03
Performance logging for MVT

lbam:


src/dso_api/dynamic_api/views/mvt.py
```diff
@@ -111,7 +115,17 @@ def setup(self, request, *args, **kwargs):
...
         kwargs.pop("table_name")
-        return super().get(request, *args, **kwargs)
+
+        t0 = datetime.now()
+        result = super().get(request, *args, **kwargs)
+        logging.info(
+            "retrieved tile for %s (%d bytes) in %s",
+            request.path,
+            len(result.content),
+            datetime.now() - t0,
```
gkoller, line 125: This will result in a negative duration if this code is executed during the transition of DST to standard time.

It's unlikely to happen, and has no real consequences, but ....

People tend to copy code and will take this as an example to do time measurement. Hence I suggest you use something like [time.perf_counter_ns](https://docs.python.org/3/library/time.html#time.perf_counter_ns).

You might even want to wrap it up in a contextmanager.

gkoller: I hate suggestion without code samples. So here it is:

```python
import time
from datetime import timedelta
from typing import Optional, Type


class Timer:
    start: int
    end: int
    duration: int

    def __init__(self) -> None:
        self.start = 0
        self.end = 0
        self.duration = 0

    def __enter__(self) -> "Timer":
        self.start = time.perf_counter_ns()
        return self

    def __exit__(
        self,
        _exc_type: Optional[Type[BaseException]],
        _exc_val: Optional[BaseException],
        _exc_tb: Optional[BaseException],
    ) -> None:
        self.end = time.perf_counter_ns()
        self.duration = self.end - self.start

    @property
    def elapsed(self) -> timedelta:
        return timedelta(microseconds=self.duration / 1_000)
```
Now your code becomes:

```python
with Timer() as t:
   result = super().get(request, *args, **kwargs)

logging.info(
    "retrieved tile for %s (%d bytes) in %s",
    request.path,
    len(result.content),
    t.elapsed)
```

Now if people start copying code, they copy something more robust and easier to use.

gkoller: Oh and to facilitate reuse even better you should add docstrings to the `Timer` class and the `elapsed` property method.

jjmurre: Or, use https://pypi.org/project/timer/ :-)

gkoller: Didn't know about that project. Nice.

What mine implementation has going for it are the type annotations. 😉

gkoller: > Or, use https://pypi.org/project/timer/ :-)

Rather coincidentally it seems that https://pypi.org/project/more-ds/ (`0.0.5`) seems to have the exact same `Timer` class that I wrote in my suggestion in https://github.com/Amsterdam/dso-api/pull/284#discussion_r645372808 (though it does away with `self.end` that is not really needed).

In addition `more-ds` does use PEP 561 to expose its type annotations so that they become available to projects using it.

lbam: Thanks. I switched to perf_counter_ns, but I don't really like the context manager solution. I'd like to keep this straight-line code, since it may well grow more complicated in the future if it is as expensive as we think.

gkoller (approved):
LGTM

PR #292, opened 2021-06-14
[AB#8346] Fix combining _fields with _expand parameters.

vdboor:
This fixes combining `_fields=..` with `_expandScope`.
This is a bit of a stop-gap to quickly fix errors for dataportaal devs.

This doesn't handle the second part yet (relation.field), but at least fixes the unwanted error message. The field parsing is already written in the same style as the expand logic so it can be revised to support nested field limiting too.


jjmurre (approved):
LGTM

PR #296, opened 2021-06-16
[AB#6571] Fix HCBRK remote client and add tests

lbam:
It was completely broken, but I didn't notice because I was running the integration tests with the wrong envvars and didn't see that they were being skipped. The endpoint isn't in use because its authorization code isn't ready yet.


jjmurre (approved):
Ziet er goed uit.

PR #297, opened 2021-06-17
Log all requests to remote endpoints, with the response status

lbam:


src/dso_api/dynamic_api/remote/clients.py
```diff
@@ -62,18 +61,19 @@ def call(self, request, path="", query_params={}) -> Union[dict, list]:  # noqa:
...
+            logger.error("Proxy call to %s failed, timeout from remote server: %s", url, e)
             raise GatewayTimeout() from e
         except (OSError, urllib3.exceptions.HTTPError) as e:
             # Socket connect / SSL error (HTTPError is the base class for errors)
-            logger.error("Proxy call failed, error when connecting to server: %s", e)
+            logger.error("Proxy call to %s failed, error when connecting to server: %s", url, e)
             raise ServiceUnavailable(str(e)) from e
 
+        level = logging.ERROR if response.status >= 500 else logging.INFO
+        logger.log(level, "Proxy call to %s, status %s: %s", url, response.status, response.reason)
```
gkoller, line 72: I would have expected all errors to be of level `logging.ERROR`

And will there even be a `response.status` set for all possible caught exceptions?

gkoller: Oh, I now see this was existing code that was simply moved. Still the questions remain valid.

lbam: The exceptions are re-raised. When we get here, we have a response, so we should also have a status.

gkoller (approved):

PR #299, opened 2021-06-20
Some various minor (layout) changes

cvriel:
- Sorting datasets in alphabetical order for easy lookup
- Sorting tables (within a dataset) in alphabetical order for easy lookup
- Sorting fields (within a table) in alphabetical order for easy lookup (where identifier fields are first)
- Mention the needed authorization (if applicable) on dataset, table and field level.

- [x] Customer friendly documentation added into `docs/`.

NA

NA


docs/source/datasets.py
```diff
@@ -46,10 +46,67 @@
 }
 
 
+def sort_schemas(schemas: Tuple[str, DatasetSchema]) -> List[Union[str, DatasetSchema]]:
+    """Sort datasets (schemas) alphabetically.
+
+    Args:
+        schemas: A set containing a pair of schemas id (name) and DatasetSchema instance.
```
jjmurre, line 53: According to the type of the argument `set` should be called `tuple` to avoid confusion.

docs/source/datasets.py
```diff
@@ -46,10 +46,67 @@
...
+
+    Returns:
+        A list of alphabetically ordered schemas instances based on their schema id (name).
+    """
+    return sorted(schemas, key=lambda schema: schema[0])
```
jjmurre, line 58: When `sorted` can be applied to the incoming argument `schemas`, that suggests it is a list (or some other kind of iterable).

jjmurre: I think input and output type should be `List[Tuple[str, DataseSchema]]`.

docs/source/datasets.py
jjmurre: A list of tuples, to be exact.

docs/source/datasets.py
cvriel: You are correct, good one

docs/source/datasets.py
cvriel: :+1:

PR #300, opened 2021-06-22
Use the identifier from the table_schema for temporality

jjmurre:
The temporal info has been move down from dataset to table.
The main temporal identifier, was still obtained from the dataset
that has now been corrected.

### DIFF
--- a/src/dso_api/dynamic_api/fields.py
+++ b/src/dso_api/dynamic_api/fields.py
@@ -3,6 +3,7 @@ from datetime import datetime, timedelta
 import azure.storage.blob
 from django.conf import settings
 from more_ds.network.url import URL
+from more_itertools import first
 from rest_framework import serializers
 from rest_framework.reverse import reverse
 from schematools.contrib.django.models import DynamicModel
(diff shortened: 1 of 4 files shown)


src/dso_api/dynamic_api/fields.py
```diff
@@ -58,9 +58,8 @@ def to_representation(self, value: DynamicModel):
 
         if href and value.is_temporal():
             table_schema = value.table_schema()
-            dataset_schema = value.get_dataset_schema()
             temporal_fieldname = table_schema.temporal.identifier
-            id_fieldname = dataset_schema["identifier"]
+            id_fieldname = table_schema.identifier[0]
```
gkoller, line 62: I'm always a little confused when I see the first element of a list explicitly referenced (eg: `[0]`). I those circumstances I ask myself the questions:
1. Will there always be only one element that just happens to be part of a list?
2. Can there be more, but are we only interested in the first element?
There is no way to make that distinction when using a plain `[0]`. However with the [more-itertools](https://more-itertools.readthedocs.io/en/stable/) library there is.

In case 1. you can use [one](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.one). In case 2. you can use [first](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.first).

I find it expressed intention much better and make code easier to read and understand.

gkoller (approved):

src/dso_api/dynamic_api/fields.py
jjmurre: Good point, I also cringe a little when I see an index-0 :-) In this case, `table_schema.identifier` is always a list and always has at least one item. But, I like the `one` and `first` functions, because, as you say, the express intention. The `more-itertools` is a great library, but I did not know these functions.

PR #301, opened 2021-07-01
AB#8005 Fix remote client error handling and logging

lbam:
Log client errors at error level.

Remove MKS header handling, it's not longer in use.

Return 401 or 403 from remote as 403. We previously returned 403 as 401 (by mistake) and returned 401 as 502 Bad Gateway.

### DIFF
--- a/src/dso_api/dynamic_api/remote/clients.py
+++ b/src/dso_api/dynamic_api/remote/clients.py
@@ -15,7 +15,7 @@ import urllib3
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from more_ds.network.url import URL
-from rest_framework.exceptions import NotAuthenticated, NotFound, ParseError, PermissionDenied
+from rest_framework.exceptions import NotFound, ParseError, PermissionDenied
 from schematools.types import DatasetTableSchema
 from urllib3 import HTTPResponse
 
@@ -211,7 +211,7 @@ class AuthForwardingClient(RemoteClient):
         if 300 <= response.status <= 399 and (
             "/oauth/authorize" in response.headers.get("Location", "")
         ):
-            raise NotAuthenticated("Invalid token")
+            raise PermissionDenied("Invalid token")
 
 
 class HCBRKClient(RemoteClient):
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/remote/clients.py
```diff
@@ -165,9 +161,11 @@ def _raise_http_error(self, response: HTTPResponse) -> None:
                 )
             else:
                 raise BadGateway(detail_message)
-        elif response.status == 403:  # "forbidden"
-            # Return 403 to client as well
-            raise NotAuthenticated(detail_message)
+        elif response.status in (401, 403):  # "unauthorized", "forbidden"
```
ngaranko, line 164: no tests for this? :(

lbam: Working on it.

lbam: Done!

gkoller: I've have suggested using [HTTPStatus](https://docs.python.org/3.8/library/http.html?highlight=httpstatus#http.HTTPStatus) before. Tried to look it up, but I can't find anything in Github these days. Anyway, I seem to remember you didn't want to use it at the time as existing code already used literal integer values. You were though, open to using `HTTPStatus` but rather would do so in a separate PR

However here you are using integer values with their meaning in comments explained. Is that really better then just using:
```python
elif response.status in (HTTPStatus.UNAUTHORIZED, HTTPStatus.FORBIDDEN):
```
Even if all other code still uses literal integer values? I mean you have to start somewhere.

lbam: I'm willing to switch, but then I will express a preference for rest_framework.status codes. Those have the numerical code in them, and while debugging this numerical codes are easier to look up and remember when you have to distinguish between NotAuthenticated, Unauthorized, Forbidden and PermissionDenied.

gkoller: So do the HTTPStatus emus.

lbam: I mean in the name: rest-framework has `status.HTTP_403_FORBIDDEN`, standard lib has `HTTPStatus.FORBIDDEN`. I want to see the status code in the source code, because that's what I'm going to grep in the RFCs :)

lbam: See updated PR.

ngaranko (approved):

gkoller (approved):

PR #304, opened 2021-07-05
Yrob/ab5522 separate path from id

Yrob:

Construct dataset api URls using the `path` property of Dataset, that is introduced in the new version of schema-tools.
This replaces earlier logic based on datasetId.


jjmurre (approved):
Ziet er goed uit, ik zie geen gekke dingen.

lbam (approved):
Nothing crazy and all tests pass locally.

PR #307, opened 2021-07-08
Refactor authentication to use native schema objects (DSO-API part)

vdboor:
This uses the new schematools authorisation framework (more info there). Instead of basing authorisation on top of the exported data in Django models, the low-level schema objects are queried by a single integrated rule engine. The UserScopes object handles both the "auth" and "profile" rules.

This also removes the need to match on field-names. Instead the model fields have a direct reference back to the schema field.

Likewise, get_fields() can exclude fields directly now and apply transformations on a field-level instead of manipulating the dict afterwards.

### DIFF
--- a/src/dso_api/dynamic_api/permissions.py
+++ b/src/dso_api/dynamic_api/permissions.py
@@ -1,107 +1,31 @@
-from dataclasses import dataclass, field
-from typing import Dict, List, Set, Type
+from typing import List
 
-from cachetools.func import ttl_cache
-from django.contrib.auth.models import AnonymousUser, _user_has_perm
-from django.db.models import Model
+from django.core.exceptions import PermissionDenied
 from rest_framework import permissions
-from rest_framework.exceptions import PermissionDenied
 from rest_framework.viewsets import ViewSetMixin
-from schematools.contrib.django import models
-from schematools.utils import to_snake_case, toCamelCase
+from schematools.permissions import UserScopes
 
 from rest_framework_dso.embedding import EmbeddedFieldMatch
 
 
-@dataclass
-class TableScopes:
-    """OAuth scopes for tables and fields."""
-
-    # Table and dataset scopes.
-    table: Set[str] = field(default_factory=set)
-    # Scope per field.
-    fields: Dict[str, str] = field(default_factory=dict)
-
-
-@ttl_cache(maxsize=None, ttl=60 * 60)
-def fetch_scopes_for_dataset_table(dataset_id: str, table_id: str):
-    """ Get the scopes for a dataset and table, based on the Amsterdam schema information """
-
-    # Make sure that names are snake_cased
-    # if used for a lookup in models.DatasetTable.objects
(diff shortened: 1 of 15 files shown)


lbam:
Looks way cleaner than the old code.

src/dso_api/dynamic_api/permissions.py
```diff
@@ -1,99 +1,25 @@
...
-        return expanded_fields
-
     result = []
     for match in expanded_fields:
-        scopes = fetch_scopes_for_model(match.field.related_model)
-        if not request.is_authorized_for(*scopes.table):
+        field_schema = match.field.field_schema
+        related_table = match.field.related_model.table_schema()
+
+        if not user_scopes.has_field_access(field_schema) or (
```
ngaranko, line 20: Maybe use `not any(user_scopes.has_field_access(field_schema), user_scopes.has_table_access(related_table))` double negative logic is harder to grasp and often leads to logical mistakes in future.

lbam: `any` needs an iterable though. We'd need to evaluate both checks. We can do

```python
not (user_scopes.has_field_access(field_schema) and user_scopes.has_table_access(related_table))
```

ngaranko: Brainfart code ;)

```
not any([user_scopes.has_field_access(field_schema), user_scopes.has_table_access(related_table)])
```

Code in PR requires either of permissions to access field, one @lbam suggesting will require both.

lbam: The check is `not A or not B`, which is `not (A and B)` or `not all([A,B])` or `any([not A, not B])`.

What I think might be more readable is to do both checks anyway, just to get the logical formula on a single line:

```python
field_access = user_scopes.has_field_access(match.field.field_schema)
table_access = user_scopes.has_table_access(match.field.related_model.table_schema())

if not (field_access and table_access):
    # ...
```

vdboor: Or something like this?

```
    for match in expanded_fields:
        field_perm = user_scopes.has_field_access(match.field.field_schema)
        table_perm = user_scopes.has_table_access(match.field.related_model.table_schema())
        if field_perm and table_perm:
            # Only add the field when there is permission
            result.append(match)
        else:
```

I dislike creating a temporary list with `all([...])` 

I'm also fine with `if not field_perm or not table_perm` as that was quite clear in my idea.

ngaranko: The less negative logic - the better imo.

lbam: Fine with me!

vdboor: aangepast!

ngaranko (approved):

lbam (approved):

jjmurre (approved):
Huge improvement, very nice piece of work!

PR #309, opened 2021-07-13
docs: update documentation on auth

vdboor:
Update the documentation based on the auth refactor.
Also simplified some wording.

### DIFF
--- a/dev-docs/source/auth.rst
+++ b/dev-docs/source/auth.rst
@@ -11,8 +11,16 @@ which are tested by the
 `datapunt-authorization-django <https://github.com/Amsterdam/authorization_django>`_
 package.
 
-Authorization
--------------
+Authorization Rulesets
+----------------------
+
+There are two mechanisms for authorization on schema level:
+
+* The ``auth`` fields in the Amsterdam Schema restrict access to resources.
+* The profiles grant permissions, which were restricted by the schema.
+
+Schema Files
+~~~~~~~~~~~~
 
 The schema definitions can add an ``auth`` field on various levels:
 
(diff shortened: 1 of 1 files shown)


dev-docs/source/auth.rst
```diff
@@ -95,18 +97,32 @@ gain permission to read the field ``bsn`` on the table ``ingeschrevenpersonen``,
...
-For example, a profile might express that a frontend office employee
-may only access data of someone when they can provide their last name and postal code.
-Or a statistician might be allowed to read age and neighbourhood
-fields to aggregate data, without ever having access to identifiable data.
-Encoding such rules in the schema file using a custom scope
-would require every other request that accesses the dataset to also have this scope.
+The ``mandatoryFilterSets`` ensures that listings are restricted on a need-to-know basis.
+Only when some information can be provided, the API grants access to see the remaining data.
+For example, a frontend office employee may only see data of someone when they can already
+provide their last name and postal code. Or a statistician might be allowed to read age and
```
mtrimpe, line 103: The statistician example isn't context dependent; right? It could saythat the statistician only gets access to certain neigbourhoods e.g. but we don't have that functionality yet so that would be confusing. 

I think what we want to say here though is that the profiles can also be used to grant access to rows and columns similarly to the 'auth' field but without cluttering the main schema needlessly... 

Or do we want to use that example to explain the randomization/first letters only feature?

dev-docs/source/auth.rst
```diff
@@ -11,50 +11,53 @@ which are tested by the
...
-The effect on a field is that the field is omitted when from the response
-when the table is queried.
-The Mapbox Vector Tiles (MVT) endpoint is an exception to this rule:
-since the MVT format requires a geometry field,
-the endpoint cannot produce valid MVT when access to the geometry field is prohibited
-and gives 403 in this case.
+The absence of an ``auth`` field makes a resource publicly available.
+
+At every level, the ``auth`` field contains a list of *scopes*.
+The JWT token of the request must contain one of these scopes in its ``scopes`` claim to access the resource.
```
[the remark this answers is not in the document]
lbam: Maybe just say that the token has to come from iam.amsterdam.nl?

dev-docs/source/auth.rst
```diff
@@ -11,50 +11,53 @@ which are tested by the
...
+
+When there is a scope at both the dataset, table and field level
+these should *all* be satisfied to have access to the field.
+
+The scopes of a dataset or table act as mandatory access control.
+When those scopes can't be satisfied, the API returns an HTTP 403 Forbidden error.
+The fields act a bit different: when the scope of a field is not satisfied,
+the field be omitted from the response.
```
lbam, line 42:
```suggestion
the field is omitted from the response.
```

dev-docs/source/auth.rst
vdboor: Good point, I've updated this part.

dev-docs/source/auth.rst
vdboor: tx!

dev-docs/source/auth.rst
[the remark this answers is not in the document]
vdboor: I'll just state "The JWT token of the request must contain one of these scopes to access the resource."

dev-docs/source/auth.rst
```diff
@@ -11,50 +11,53 @@ which are tested by the
...
+
+Sometimes it's not possible to remove a field (for example, a geometry field for Mapbox Vector Tiles).
+In such case, the endpoints produces a HTTP 403 error to deny complete access.
```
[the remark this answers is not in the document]
vdboor: tx!

dev-docs/source/auth.rst
```diff
@@ -95,18 +97,36 @@ gain permission to read the field ``bsn`` on the table ``ingeschrevenpersonen``,
...
+Application in DSO-API
+----------------------
+
+The dataset and profile files stored in the repository for Amsterdam Schema.
+Both are imported into the DSO-API database, and loaded once on startup.
+
+Schematools
+~~~~~~~~~~~
+
+The authorization engine is implemented within ``schematools`` as low-level Python object.
```
lbam, line 119:
```suggestion
The authorization engine is implemented within ``schematools`` as low-level Python objects.
```

lbam:
Looks like I hit the button that marks review comments as "pending" instead of just posting them. Sorry about that.

dev-docs/source/auth.rst
```diff
@@ -95,18 +97,36 @@ gain permission to read the field ``bsn`` on the table ``ingeschrevenpersonen``,
...
+~~~~~~~~~~~
+
+The authorization engine is implemented within ``schematools`` as low-level Python objects.
+The ``UserScopes`` class provides the main logic, which is accessed within the DSO-API
+as ``request.user_scopes.has_..._access()``. Each access function returns a
+:class:`~schematools.types.Permission` object with the granted access level.
+When no permission is given, the object evaluates to ``False`` in boolean comparisons (e.g. ``if permission``).
 
+The :class:`~schematools.types.Permission` object provides a ``level``, ``sub_value`` and ``transform_function()``
+for fine granted access levels, such as only viewing a field as encoded or only it's first 3 letters.
```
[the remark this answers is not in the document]
vdboor: Fixed in dac3902b03cdf11fef079c3c589a07b0960972b5

vdboor:
> Looks like I hit the button that marks review comments as "pending" instead of just posting them. Sorry about that.

no problem, fixed in dac3902b03cdf11fef079c3c589a07b0960972b5.

PR #310, opened 2021-07-13
API docs: order of tables are presented by order of appearance

cvriel:
On request of datateam omgevingswet the presentation order of the tables within a dataset is not ordered on alphabet but on appearance within the Amsterdam schema specification. The order of the tables defined in the Amsterdam schema is based on a certain logical and intentional meaning. In respect of the self defined order of tables the datateams can use the docs more intuatieve.

v1/docs/

N.A.

N.A.

N.A.


jjmurre (approved):
👍🏼

PR #311, opened 2021-07-15
No alphabetical sort order for fields.

cvriel:
In addition to removing the alphabetical sort order for tables (within a dataset), also the alphabetical sort order for fields is removed. So it orders by the definition order in Amsterdam Schema. So the order is in line with the initial own logical order that the datateams specified in the schema.

v1/docs/

N.A.

N.A.

N.A.


jjmurre (approved):
LGTM

PR #317, opened 2021-07-29
Restore remote authorization

lbam:
Remote authorization is restored (currently, it's left entirely to the remote). Results of authorization checks are logged. Connection to HC-BRK is not yet established, but the parts are in place. In `tests/files`, there's a preliminary version of the detailed HC-BRK authorizing schema.


lbam:
The commits should be squashed. The first commit actually reverts an earlier change because I screwed up a rebase...

PR #319, opened 2021-08-05
Fix build serializer field for AutoField

Yrob:
_build_serializer_field called .get_field_schema() on AutoFields.
This caused a crash on startup. This commit fixes that bug.


ngaranko (approved):
Would be nice to have a test for this.

PR #321, opened 2021-08-18
Yrob/11742 breadcrumbs

Yrob:
Breadcrumbs zijn nu:
`DSO-API Datasets/<dataset-naam>/<tabel-naam>/<instance-title>`
Of voor datasets op een sub path:
`DSO-API Datasets/<dataset-path>/<dataset-naam>/<tabel-naam>/<instance-title>`

Ook de header in browser view is geupdate. Daar staat nu niet meer "Django Rest Framework", maar "Amsterdam Datadiensten"
met het logo van Gemeente Amsterdam.


src/dso_api/dynamic_api/routers.py
```diff
@@ -48,14 +48,21 @@
     from schematools.contrib.django.models import DynamicModel
 
 
-class DynamicAPIRootView(APIIndexView):
+class DynamicAPIIndexView(APIIndexView):
```
ngaranko, line 51: There's no need to rename this class, it is still Root view.

src/rest_framework_dso/views.py
```diff
@@ -20,6 +21,35 @@
...
+    This name is used in the browsable API, and in OPTIONS responses.
+
+    This function repaces the Rest Framework default.
+    """
+    name = getattr(view, "name", None)
+    if name is not None:
+        return name
+
+    name = view.__class__.__name__
+    name = formatting.remove_trailing_string(name, "View")
```
jjmurre, line 36: Nice, didn't know these util functions.

src/dso_api/dynamic_api/routers.py
```diff
@@ -250,6 +262,35 @@ def _build_openapi_views(self, datasets: Iterable[Dataset]) -> List[URLPattern]:
...
+            paths += ds_paths[0:-1]
+        paths = set(paths)
+
+        # Create an index view for each path
+        results = []
+        for p in paths:
+            name = p.split("/")[-1].capitalize() + " Datasets"
+            results.append(
+                path(
+                    p + "/",
```
ngaranko, line 287: What if multiple datasets are using same subpath? How this will look?

ngaranko (commented):
Please add tests for every change in this PR.

src/templates/dso_api/dynamic_api/api.html
```diff
@@ -0,0 +1,70 @@
+{% extends "rest_framework/base.html" %}
```
jjmurre, line 1: Brand building :-)

jjmurre (approved):
LGTM

src/dso_api/dynamic_api/routers.py
Yrob: Like the index page at root, but with only the datasets on that path

src/dso_api/dynamic_api/routers.py
Yrob: But its now also an index view at locations other than root

ngaranko: Those are roots of subtrees, right? :)

Yrob: So it should be called DynamicAPINodeView? or DynamicAPISubTreeView :-)

Yrob:
> Please add tests for every change in this PR.

Done

src/dso_api/dynamic_api/routers.py
ngaranko: Why not keep DynamicAPIRootView? It's just fine.

src/dso_api/dynamic_api/routers.py
ngaranko: This is still not covered by tests.

Yrob: Doesnt the new test in  src/tests/test_openapi.py cover this? It has two datasets on the same subpath and checks both are listed.

ngaranko: Oh, it does, missed.

PR #322, opened 2021-08-19
Breadcrumbs work with http errors

Yrob:
If an instance view returns a 404 or other http error
this will now display nicely in the page title and breadcrumbs

### DIFF
--- a/src/rest_framework_dso/views.py
+++ b/src/rest_framework_dso/views.py
@@ -41,12 +41,15 @@ def get_view_name(view):
     if name == "Schema":
         name = getattr(view.schema_generator, "title", "Schema")
 
-    # Add the title for instances
+    # Use the title for instances
     response = getattr(view, "response", None)
     suffix = getattr(view, "suffix", None)
 
     if suffix == "Instance" and response:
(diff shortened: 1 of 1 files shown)


src/rest_framework_dso/views.py
```diff
@@ -41,12 +41,15 @@ def get_view_name(view):
...
         name = getattr(view.schema_generator, "title", "Schema")
 
-    # Add the title for instances
+    # Use the title for instances
     response = getattr(view, "response", None)
     suffix = getattr(view, "suffix", None)
 
     if suffix == "Instance" and response:
-        name = response.data["_links"]["self"]["title"]
+        if response.status_code == 200:
```
jjmurre, line 49: Nice solution!

jjmurre (approved):
LGTM

src/rest_framework_dso/views.py
Yrob: thanks

ngaranko: Please use `rest_framework.status. HTTP_200_OK`. Also adding test is good idea.

PR #323, opened 2021-08-20
AB#5522 mvt and wfs views on sub paths

Yrob:
The mvt and wfs views were not using the dataset.path
property yet. So datasets with subpaths appeared on
their old paths. This update fixes that.

MVT and WFS url pattern logic is moved from get_patterns
into the custom router. Because get_patterns doesnt work with
nested paths.

### DIFF
--- a/src/dso_api/dynamic_api/routers.py
+++ b/src/dso_api/dynamic_api/routers.py
@@ -23,6 +23,7 @@ the :func:`~schematools.contrib.django.factories.model_factory`,
 from __future__ import annotations
 
 import logging
+from itertools import chain
 from typing import TYPE_CHECKING, Dict, Iterable, List, Type
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


src/tests/test_openapi.py
```diff
@@ -128,8 +128,8 @@ def test_subpath_view(
                     }
                 ],
                 "related_apis": [
-                    {"type": "WFS", "url": f"{base}/v1/wfs/afvalwegingen/"},
-                    {"type": "MVT", "url": f"{base}/v1/mvt/afvalwegingen/"},
+                    {"type": "WFS", "url": f"{base}/v1/wfs/sub/path/afvalwegingen/"},
```
gkoller, line 131:
```python
from more_ds.network import URL
BASE = URL(base)
...
{"type": "WFS", "url": BASE / "v1/wfs/sub/path/afvalwegingen/"},
...
```
Not a whole lot of benefit character wise but having one, and only one consistent way of URL construction would be beneficial.

src/dso_api/dynamic_api/routers.py
```diff
@@ -135,7 +142,13 @@ def initialize(self):
...
     def get_urls(self):
         """Add extra URLs beside the registered viewsets."""
-        return super().get_urls() + self._openapi_urls + self._index_urls
+        return (
+            super().get_urls()
+            + self._openapi_urls
+            + self._index_urls
+            + self._mvt_urls
+            + self._wfs_urls
+        )
```
gkoller, line 151: This is not performance critical code but instead of creating 5 separate lists, only to throw away 4 of them feels not very Pythonic to me. I'd rather see:

```python
from itertools import chain

return list(chain(
    super().get_urls(),
    self._openapi_urls,
    self._index_urls,
    self._mvt_urls,
    self._wfs_urls))
```

gkoller (dismissed):

PR #328, opened 2021-08-30
Fix nested expand support, was broken after 2 levels

vdboor:
After increasing the number of allowed embedding, several bugs showed up. This PR also fixes those bugs.

* Added `ExpandScope` object that retains the nested embeds. Changed `fields_to_expand` to `expand_scope`, since the nested expand was lost by only passing `.keys()` to the sub-level serializer.
* Replaced `lru_cache()` with cachetools for `serializer_factory()`. This way the `nesting_level` no longer changes caching.
* Fixed recursion of the `serialize_factory()` for self-references, which avoided the LRU-cache.
* Completed BAG models for unit tests, so deep nesting could be tested.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -633,9 +633,9 @@ def _build_serializer_field(
     model_field: models.Field,
     flat: bool,
     nesting_level: int,
-    new_attrs,
-    fields,
-    extra_kwargs,
+    new_attrs: dict,
+    fields: list,
+    extra_kwargs: dict,
 ):
     """Build a serializer field, results are written in 'output' parameters"""
     # Add extra embedded part for related fields
@@ -650,7 +650,7 @@ def _build_serializer_field(
         ),
     ):
         # Embedded relations are only added to the main serializer.
-        _build_serializer_embeddded_field(model_field, nesting_level, new_attrs)
+        _build_serializer_embedded_field(model_field, nesting_level, new_attrs)
 
         if isinstance(model_field, LooseRelationField):
             # For loose relations, add an id char field.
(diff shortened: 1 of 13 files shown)


src/rest_framework_dso/utils.py
```diff
@@ -0,0 +1,10 @@
+from django.utils.functional import LazyObject, empty
+
+
+def unlazy_object(obj):
```
jjmurre, line 4: Clever tricks with the lazy object! Are you not afraid the could break, because it uses Django internals?

vdboor: No not really, it's covered by tests and a pretty stable part of Django.

src/dso_api/dynamic_api/serializers.py
```diff
@@ -836,7 +836,8 @@ def _find_reverse_fk_relation(
     )
 
 
-def _generate_embedded_relations(model, fields, new_attrs, nesting_level):
+def _generate_nested_relations(model, fields, new_attrs, nesting_level):
```
jjmurre, line 866: Sprinkle in some types here and there.

src/dso_api/dynamic_api/serializers.py
```diff
@@ -474,10 +476,18 @@ def to_representation(self, obj):
...
+serializer_factory_cache = LRUCache(maxsize=100000)
+dynamic_models_removed.connect(lambda **kwargs: serializer_factory_cache.clear())
 
 
-@lru_cache()
+def _serializer_cache_key(model, depth=0, flat=False, nesting_level=0):
+    """The cachetools allow definining the cache key,
+    so the nesting level does not bypass the cache which it would with lru_cache().
+    """
+    return hashkey(model, depth, flat)
```
jjmurre, line 487: Nice, didn't you could get this fine-grained control on parameters.

jjmurre (dismissed):
Nice addition to the DSO!

src/dso_api/dynamic_api/serializers.py
vdboor: done :-)

jjmurre (approved):
Look good!

jjmurre (approved):
OK

PR #336, opened 2021-09-09
AB#6554 Don't leak responses from remotes to unauthorized clients

lbam:
Hotfix. Will restore a sanitized version of the detailed error in a later commit.

### DIFF
--- a/src/dso_api/dynamic_api/remote/views.py
+++ b/src/dso_api/dynamic_api/remote/views.py
@@ -102,11 +102,7 @@ class RemoteViewSet(DSOViewMixin, ViewSet):
         if not serializer.is_valid():
             raise RemoteAPIException(
                 title="Invalid remote data",
-                detail={
-                    "detail": "These schema fields did not validate:",
-                    "x-validation-errors": serializer.errors,
-                    "x-raw-response": raw_data,
-                },
+                detail="Some schema fields did not validate",
                 code="validation_errors",
                 status_code=status.HTTP_502_BAD_GATEWAY,
             )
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/remote/views.py
```diff
@@ -102,11 +102,7 @@ def validate(self, serializer, raw_data):
         if not serializer.is_valid():
             raise RemoteAPIException(
                 title="Invalid remote data",
-                detail={
-                    "detail": "These schema fields did not validate:",
-                    "x-validation-errors": serializer.errors,
-                    "x-raw-response": raw_data,
-                },
+                detail="Some schema fields did not validate",
```
jjmurre, line 105: We could log this information for debugging purposes. Although, when there is private info in there, that should also not end up in the logs. So, better like this.

jjmurre (approved):
LGTM

src/dso_api/dynamic_api/remote/views.py
ngaranko: Logging this information is bad indeed. But maybe logging only keys from `raw_data` to error log would be enough for debugging?

lbam: Good point.

ngaranko: Cool, why this was merged then?

lbam: Because this is a hotfix. Next PR will have a more thorough solution.

PR #337, opened 2021-09-09
AB#6554 Log sanitized version of remote serializer errors

lbam:
Follow up to #336.


lbam:
Should we still put the field names (`serializer.errors.keys()`) in our reponse's `detail`?

Yrob (approved):

PR #342, opened 2021-09-15
AB#6554 Remove HCBRK integration test

lbam:
This test is outdated. I tried to fix it, but ultimately decided that testing in the browser is easier and testing the production environment is more important. We need different kinds of tests for that. Jenkins doesn't run this test anyway.


jjmurre (approved):
Indeed, better to remove it.

PR #346, opened 2021-09-16
Guido/python39

gkoller:
Upgrade to Python 3.9

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
-        language_version: python3.8
+        language_version: python3.9
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.4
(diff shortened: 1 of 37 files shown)


.pre-commit-config.yaml
```diff
@@ -6,7 +6,6 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
-        language_version: python3.8
```
vdboor, line 9: Niet python3.9?

vdboor (approved):

.pre-commit-config.yaml
gkoller: Was about to say: _Nope. By default `black` uses the version of the Python interpreter it is running on_ But that only holds when you are running `black` from the command line in the project's Python environment. I don't now how `pre-commit` selects the Python interpreter for its own environment. So yes, we should specify it. Good catch :+1:

PR #349, opened 2021-09-20
Refactor serializer factory logic to avoid passing many out-variables to functions

vdboor:
Instead collect everything in a single SerializerAssemblyLine class.
This also streamlines how both 'fields' and 'new_attrs' has to be set.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -487,6 +487,59 @@ def _serializer_cache_key(model, depth=0, flat=False, nesting_level=0):
     return hashkey(model, depth, flat)
 
 
+class SerializerAssemblyLine:
+    """The intermediate result of the serializer factory"""
+
+    def __init__(
+        self,
+        model: type[DynamicModel],
+        fields=None,
+        depth: int = 0,
+        openapi_docs: str = "",
+        **meta_kwargs,
+    ):
+        """
+        :param model: The model for which this serializer is created.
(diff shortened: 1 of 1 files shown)


src/dso_api/dynamic_api/serializers.py
```diff
@@ -512,26 +558,23 @@ def serializer_factory(
...
-        "__module__": f"dso_api.dynamic_api.serializers.{safe_dataset_id}",
-        "__doc__": table_schema.get(
-            "description", table_schema.name
-        ),  # avoid exposing our docstrings.
-    }
+
+    serializer_part = SerializerAssemblyLine(model, depth=depth)
+
+    # Generate a proper docstring for the OpenAPI spec
+    serializer_part.class_attrs["__doc__"] = table_schema.get("description", table_schema.name)
```
jjmurre, line 571: Why not doing this inside `SerializerAssemblyLine`? I saw the comment about not exposing the docstring. Does doing it like this avoid exposing the docstring?

jjmurre (approved):
Nice refactoring.

src/dso_api/dynamic_api/serializers.py
vdboor: That class is reused for more serializer construction, including the `_links` serializer and through serializer. I didn't want to render those docs for these classes as well.

vdboor: I've moved this to a constructor argument

PR #351, opened 2021-09-21
Log response times from remotes

lbam:
To debug the response time issues with HCBRK.

### DIFF
--- a/src/dso_api/dynamic_api/remote/clients.py
+++ b/src/dso_api/dynamic_api/remote/clients.py
@@ -6,6 +6,7 @@ to avoid the overhead of the requests library.
 
 import logging
 import threading
+import time
 from typing import Union
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


lbam:
The 's' is for seconds. It's not interpreted by the formatter (I checked).

src/dso_api/dynamic_api/remote/clients.py
```diff
@@ -70,7 +72,14 @@ def call(self, request, path="", query_params={}) -> Union[dict, list]:  # noqa:
...
         level = logging.ERROR if response.status >= 400 else logging.INFO
-        logger.log(level, "Proxy call to %s, status %s: %s", url, response.status, response.reason)
+        logger.log(
+            level,
+            "Proxy call to %s, status %s: %s, took: %.3fs",
+            url,
+            response.status,
+            response.reason,
+            (time.perf_counter_ns() - t0) * 1e-9,
+        )
```
gkoller, line 82: This way you'll only be logging successful calls. I take it time spend making unsuccessful calls is equally valuable?

lbam: I'm hoping `urllib3.exceptions.TimeoutError` has the timeout in the error message, though TBH, I didn't check.

PR #358, opened 2021-09-24
Fix recursion in OpenAPI when objects reference each other

vdboor:


jjmurre (approved):
LGTM. Looks to me it even covers indirect recursion (A -> B -> C -> A).

PR #360, opened 2021-09-27
Implemented reverse embedding

vdboor:
The implementation is done by tracking the primary-keys in the `id_list` and query the foreign table on a foreign key. It allowed to keep the existing machinery in place, which relies on streaming data by collecting object identifier's during the rendering.

This also changes the way embedded fields can be registered. In the end it wasn't required to build this, but it's still a good feature to have.

### DIFF
--- a/src/rest_framework_dso/embedding.py
+++ b/src/rest_framework_dso/embedding.py
@@ -203,6 +203,22 @@ class EmbeddedFieldMatch:
 
         return serializer
 
+    def get_id_fetcher(self) -> Callable[[list[Union[int, str]], str], models.QuerySet]:
+        """Provide the function that allows to collect objects by ID.
+        This hook allows the serializer to override this logic,
+        e.g. to retrieve only active records.
+        """
+        if self.embedded_serializer.parent.id_based_fetcher:
+            # Use to serializer based ID-fetcher if needed.
+            return self.embedded_serializer.parent.id_based_fetcher(self.field)
+        else:
+            model = self.field.related_model
+
+            def fetcher(id_list, id_field=None):
+                return model.objects.filter(**{f"{id_field or 'pk'}__in": id_list})
+
+            return fetcher
+
 
 def get_all_embedded_field_names(
     serializer_class: type[serializers.Serializer],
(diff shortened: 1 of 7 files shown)


src/rest_framework_dso/embedding.py
```diff
@@ -273,28 +289,20 @@ def get_embedded_field(
...
+        if not embedded_fields:
+            raise ParseError(
+                f"Eager loading is not supported for field '{prefix}{field_name}'"
+            ) from None
+        else:
+            available = f", {prefix}".join(sorted(embedded_fields.keys()))
+            raise ParseError(
+                f"Eager loading is not supported for field '{prefix}{field_name}', "
+                f"available options are: {prefix}{available}"
+            ) from None
```
lbam, line 305:
```suggestion
        msg = f"Eager loading is not supported for field '{prefix}{field_name}'"
        if embedded_fields:
            available = f", ".join(prefix + k for k in sorted(embedded_fields.keys()))
            msg += f"available options are: {available}"
        raise ParseError(msg) from None
```

src/rest_framework_dso/embedding.py
```diff
@@ -481,11 +489,25 @@ class EmbeddedResultSet(ReturnGenerator):
     """A wrapper for the returned expanded fields.
     This is used in combination with the ObservableIterator.
 
+    This result set is lazy evaluated. It doesn't contain any results on creation.
```
lbam, line 492:
```suggestion
    This result set is lazily evaluated. It doesn't contain any results on creation.
```

src/rest_framework_dso/fields.py
```diff
@@ -19,12 +23,32 @@ def parse_request_fields(fields: Optional[str]):
...
+) -> type[AbstractEmbeddedField]:
+    """Return the embedded field type that is suited for a particular model field."""
+    if isinstance(model_field, ForeignObjectRel):
+        # Reverse relations
+        return EmbeddedManyToOneRelField
+    else:
+        if model_field.one_to_many or model_field.many_to_many:
+            return EmbeddedManyToManyField
+        else:
+            return EmbeddedField
```
lbam, line 37:
```suggestion
def get_embedded_field_class(
    model_field: Union[RelatedField, LooseRelationField, ForeignObjectRel]
) -> type[AbstractEmbeddedField]:
    """Return the embedded field type that is suited for a particular model field."""
    if isinstance(model_field, ForeignObjectRel):
        # Reverse relations
        return EmbeddedManyToOneRelField
    elif model_field.one_to_many or model_field.many_to_many:
        return EmbeddedManyToManyField
    else:
        return EmbeddedField
```

src/rest_framework_dso/fields.py
```diff
@@ -19,12 +23,32 @@ def parse_request_fields(fields: Optional[str]):
...
+
 class AbstractEmbeddedField:
     """A 'virtual' field that contains the configuration of an embedded field.
 
     Note this virtual field is *not* part of the serializer.fields,
     thus it's also **not** copied per instance. It's not possible to store
     request-state (such as the parent serializer object) in this class.
+
+    There are two ways to include an embedded field in the serializer.
+    Either add it as class attribute (it will auto-register
```
lbam, line 48:
```suggestion
    Either add it as a class attribute (it will auto-register
```

src/rest_framework_dso/fields.py
```diff
@@ -19,12 +23,32 @@ def parse_request_fields(fields: Optional[str]):
...
+    in ``Meta.embedded_fields``), or add it directly in ``Meta.embedded_fields``.
+    Both styles require the serializer class to extend
```
lbam, line 50:
```suggestion
    Both styles require the serializer class to inherit
```

src/rest_framework_dso/fields.py
vdboor: Good point, but tricky. I've updated the code to explicitly show why I'd taken this if/else path.
The reverse M2M likely needs to become a separate type.

lbam (approved):

PR #364, opened 2021-09-29
Implement reverse M2M embedded field and optimize querying.

vdboor:
Based on insights from reverse-embedded FK fields, the M2M implementation could be optimized. Before it queried the through table for *each* individual record to find related ID's. Now it collects a list of source ID's, and adjusts the query-filter to retrieve related objects in a single query. This also made the implementation of the reverse M2M very easy: the reverse ORM path has to be followed.

The old code also mixed up "identificatie" and "pk" fields, meaning that historical/temporal data likely wasn't returned at all. Tests also had this bug, and gave a false-negative after making changes. It turned out that the tested through table was incorrectly filled with non-existant IDs.

Additionally, repeated relations (forward/back/forward) became a problem. The serializers properly detect that the same relation object is traversed again and filter out those fields in both `_links` and the embedding.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -425,6 +425,72 @@ class DynamicLinksSerializer(DynamicSerializer):
 
     serializer_related_field = HALTemporalHyperlinkedRelatedField
 
+    def get_fields(self):
+        """Avoid adding fields that loop back to a parent."""
+        fields = super().get_fields()
+
+        if self.parent_source_fields:
+            for name, field in fields.copy().items():
+                if self._is_repeated_relation(name, field):
+                    logger.debug(
+                        "Excluded %s from _links field, it resolves to a parent via %s.%s",
+                        name,
+                        ".".join(f.name for f in self.parent_source_fields),
+                        field.source,
+                    )
+                    del fields[name]
+
+        return fields
+
+    @cached_property
+    def parent_source_fields(self) -> list[Union[models.Field, ForeignObjectRel]]:
+        """Find the ORM relationship fields that lead to this serializer instance"""
+        source_fields = []
+        parent = self.parent
+
+        # The root-level obviously doesn't have a field name,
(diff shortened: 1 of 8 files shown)


src/dso_api/dynamic_api/serializers.py
```diff
@@ -425,6 +425,74 @@ class DynamicLinksSerializer(DynamicSerializer):
...
+
+        if self.parent_source_fields:
+            for name, field in fields.copy().items():
+                if self._is_repeated_relation(name, field):
+                    logger.debug(
+                        "Excluded %s from _links field, it resolves to a parent via %s.%s",
+                        name,
+                        ".".join(f.name for f in self.parent_source_fields),
+                        field.source,
+                    )
```
lbam, line 440: The string join is executed even if the logger is set to a higher than `DEBUG`.
```suggestion
                    if logger.isEnabledFor(logging.DEBUG):
                        logger.debug(
                            "Excluded %s from _links field, it resolves to a parent via %s.%s",
                            name,
                            ".".join(f.name for f in self.parent_source_fields),
                            field.source,
                        )
```

src/dso_api/dynamic_api/serializers.py
```diff
@@ -425,6 +425,74 @@ class DynamicLinksSerializer(DynamicSerializer):
 
     serializer_related_field = HALTemporalHyperlinkedRelatedField
 
+    def get_fields(self):
```
lbam, line 428: Type annotation for the return value?

lbam (approved):

src/dso_api/dynamic_api/serializers.py
vdboor: Adressed in PR #366 :-)

src/dso_api/dynamic_api/serializers.py
vdboor: Adressed in PR #366 :-)

PR #389, opened 2021-11-03
Fixed listings and embedding to only return the right temporal records

vdboor:
Each commit holds a clear step for this PR.

* Removed lots of old code that was no longer used.
* Listings now return only active versions.
* Embeds also return active versions only.
* Test code lacked `beginGeldigheid` data, was added.
* Cleaned up test code URL construction (list vs detail)

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -318,8 +318,12 @@ class DynamicSerializer(DSOModelSerializer):
         As the field name doesn't reference the model field directly,
         DRF assumes it's an "@property" on the model.
         """
-        forward_map = model_class._meta._forward_fields_map.get(field_name)
-        if forward_map and isinstance(forward_map, models.ForeignKey):
+        model_field = model_class._meta._forward_fields_map.get(field_name)
+        if (
+            model_field is not None
+            and isinstance(model_field, models.ForeignKey)
+            and model_field.related_model.is_temporal()
+        ):
             return TemporalReadOnlyField, {}
         else:
             return super().build_property_field(field_name, model_class)
@@ -964,8 +968,7 @@ def _build_href_field(
     """Generate a Hyperlink related field, or one with temporal handling.
     Use the 'lookup_field' argument to change the source of the hyperlink ID.
     """
-    temporal: Optional[Temporal] = target_model.table_schema().temporal
-
+    table_schema = target_model.table_schema()
     href_kwargs = dict(
         view_name=get_view_name(target_model, "detail"),
         read_only=True,  # avoids having to add a queryset
@@ -973,9 +976,9 @@ def _build_href_field(
         lookup_field=lookup_field,
         **kwargs,
     )
-    if temporal is not None:
+    if table_schema.temporal is not None:
         # Temporal version that splits the ID
-        return TemporalHyperlinkedRelatedField(temporal=temporal, **href_kwargs)
+        return TemporalHyperlinkedRelatedField(table_schema=table_schema, **href_kwargs)
     else:
         return HyperlinkedRelatedField(**href_kwargs)
 
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -48,6 +48,7 @@ from dso_api.dynamic_api.fields import (
     TemporalReadOnlyField,
 )
 from dso_api.dynamic_api.permissions import filter_unauthorized_expands
+from dso_api.dynamic_api.temporal import TemporalTableQuery, filter_temporal_slice
 from dso_api.dynamic_api.utils import get_source_model_fields, resolve_model_lookup
 from rest_framework_dso.embedding import EmbeddedFieldMatch
 from rest_framework_dso.fields import (
@@ -61,20 +62,6 @@ MAX_EMBED_NESTING_LEVEL = 10
 logger = logging.getLogger(__name__)
 
 S = TypeVar("S", bound=serializers.Serializer)
-
-
-def filter_latest_temporal(queryset: models.QuerySet) -> models.QuerySet:
-    """Make sure a queryset will only return the latest temporal models"""
-    table_schema = queryset.model.table_schema()
-    temporal_config = table_schema.temporal
-    if not temporal_config:
-        return queryset
-
-    identifier = first(table_schema.identifier)
-    sequence_name = table_schema.temporal.identifier
-
-    # does SELECT DISTINCT ON(identifier) ... ORDER BY identifier, sequence DESC
-    return queryset.distinct(identifier).order_by(identifier, f"-{sequence_name}")
 
 
 @extend_schema_field(
@@ -91,13 +78,17 @@ class RelatedSummaryField(Field):
     def to_representation(self, value: models.Manager):
         request = self.context["request"]
         url = reverse(get_view_name(value.model, "list"), request=request)
+
+        # the "core_filters" attribute is available on all related managers
         filter_field = next(iter(value.core_filters.keys()))
         q_params = {toCamelCase(filter_field + "_id"): value.instance.pk}
 
-        # If this is a temporary slice, add the extra parameter to the qs.
-        if request.table_temporal_slice is not None:
-            key = request.table_temporal_slice["key"]
-            q_params[key] = request.table_temporal_slice["value"]
+        # If this is a temporal table, only return the appropriate records.
+        if value.model.is_temporal():
+            # The essence of filter_temporal_slice()
+            query = TemporalTableQuery(request, value.model.table_schema())
+            q_params.update(query.url_parameters)
+            value = query.filter_queryset(value)
 
         query_string = ("&" if "?" in url else "?") + urlencode(q_params)
         return {"count": value.count(), "href": f"{url}{query_string}"}
@@ -109,18 +100,20 @@ class DynamicListSerializer(DSOModelListSerializer):
     """
 
     def get_prefetch_lookups(self) -> list[Union[models.Prefetch, str]]:
-        """Optimize M2M prefetch lookups to return latest temporal records only."""
+        """Optimize M2M prefetch lookups to return correct temporal records only."""
         parent_model = self.child.Meta.model
         lookups = super().get_prefetch_lookups()
+        request = self.context["request"]
 
         for i, lookup in enumerate(lookups):
             related_model, is_many = resolve_model_lookup(parent_model, lookup)
             if cast(DynamicModel, related_model).is_temporal() and is_many:
 
-                # When the model is a temporal relationship, make sure the prefetch only
-                # returns the latest objects. The Prefetch objects allow adding these filters.
+                # When the model is a temporal relationship, make sure the prefetch only returns
+                # the proper temporal slice. The Prefetch objects allow adding these filters.
                 lookups[i] = models.Prefetch(
-                    lookup, queryset=filter_latest_temporal(related_model.objects)
+                    lookup,
+                    queryset=filter_temporal_slice(request, related_model.objects.all()),
                 )
         return lookups
 
@@ -271,10 +264,13 @@ class DynamicSerializer(DSOModelSerializer):
         This override makes sure the correct temporal slice is returned.
         """
         if embedded_field.is_loose:
-            # Special handling for embedding temporal relationships.
+            # Special handling for embedding temporal relationships, if they are loose relations
+            # When 'identifier' is ["identificatie", "volgnummer"], take the first as grouper.
             model = embedded_field.related_model
             id_field = embedded_field.related_id_field or model.table_schema().identifier[0]
-            return filter_latest_temporal(model.objects).filter(**{f"{id_field}__in": id_list})
+            return filter_temporal_slice(self._request, model.objects.all()).filter(
+                **{f"{id_field}__in": id_list}
+            )
         else:
             return super().get_embedded_objects_by_id(embedded_field, id_list=id_list)
 
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import logging
 from functools import wraps
-from typing import Any, Callable, Optional, TypeVar, Union, cast
+from typing import Any, Callable, Iterable, Optional, TypeVar, Union, cast
 from urllib.parse import urlencode
 
 from cachetools import LRUCache, cached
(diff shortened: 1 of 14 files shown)


src/dso_api/dynamic_api/views/api.py
```diff
@@ -51,7 +52,9 @@ def get_queryset(self):
         if self.request.versioned and self.model.is_temporal():
             if self.request.table_temporal_slice is not None:
                 temporal_value = self.request.table_temporal_slice["value"]
-                start_field, end_field = self.request.table_temporal_slice["fields"]
+                start_field, end_field = map(
```
jjmurre: Good! Now the casing is done at a better place in the stack.

src/dso_api/dynamic_api/serializers.py
```diff
@@ -286,6 +263,21 @@ def _raise_better_embed_error(cls, field_name, prefix=""):
                     f" for embedding as it's a summary of a huge listing."
                 ) from None
 
+    def get_embedded_objects_by_id(
```
jjmurre, line 259: I really like this refactor! Much cleaner now and totally in the spirit of DRF.

src/tests/test_dynamic_api/test_temporal_actions.py
```diff
@@ -136,18 +136,21 @@ def test_details_record_can_be_requested_by_pk(self, api_client, stadsdelen):
...
+        response = api_client.get(url)
         data = read_response_json(response)
 
         assert response.status_code == 200, data
         assert data["_links"]["self"]["volgnummer"] == 2, data
+        assert data["id"] == stadsdelen[1].id, data
 
     def test_details_can_be_requested_with_valid_date(self, api_client, stadsdelen):
         """Prove that object can be requested by identification and date,
         resulting in correct for that date object."""
```
jjmurre, line 166: Small nitpick:  "resulting in correct for" -> seems something is missing in that sentence.

src/tests/test_dynamic_api/test_temporal_actions.py
```diff
@@ -156,8 +159,9 @@ def test_details_can_be_requested_with_valid_date(self, api_client, stadsdelen):
     def test_details_can_be_requested_with_version(self, api_client, stadsdelen):
         """Prove that object can be requested by identification and version,
         resulting in correct for that version object."""
```
jjmurre, line 177: Same here.

src/dso_api/dynamic_api/serializers.py
vdboor: yeah moving this logic to the serializer avoided this `id_fetcher` mess and made overriding much easier. Took some time to figure that one out! :-D

src/dso_api/dynamic_api/serializers.py
```diff
@@ -286,6 +256,23 @@ def _raise_better_embed_error(cls, field_name, prefix=""):
...
+        self, embedded_field: AbstractEmbeddedField, id_list: list[Union[str, int]]
+    ) -> Union[models.QuerySet, Iterable[models.Model]]:
+        """Retrieve a number of embedded objects by their identifier.
+
+        This override makes sure the correct temporal slice is returned.
+        """
+        if embedded_field.is_loose:
+            # Special handling for embedding temporal relationships, if they are loose relations
+            model = embedded_field.related_model
+            id_field = embedded_field.related_id_field or model.table_schema().identifier[0]
```
gkoller, line 269: Re: `model.table_schema().identifier[0]` Why are you only interested in the the first element of `identifier`and not the others?

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+        self.table_schema = table_schema
+        self.is_versioned = temporal is not None
+
+        # Parsed arguments are None unless there is a match with the request
+        self.version_field = None
+        self.version_value = None
+
+        self.slice_dimension = None
+        self.slice_value = None
+        self.slice_range_fields = None
```
gkoller, line 51: You should set these default values in the class definition (where you define their types) in the spirit of PEP 526.

Eg, instead of:
```python
class TemporalTableQuery:
    version_field: Optional[str]
    def __init__(...):
        self.version_field = None
```
You should do:
```python
class TemporalTableQuery:
    version_field: Optional[str] = None
```

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+
+    #: Which date is requested
+    slice_dimension: Optional[str]
+    slice_value: Optional[Union[Literal["*"], date, datetime]]
+    slice_range_fields: Optional[TemporalDimensionFields]
+
+    def __bool__(self):
+        return self.is_versioned
+
+    def __init__(self, request, table_schema: DatasetTableSchema):
```
gkoller, line 39: `-> None`

src/dso_api/dynamic_api/views/api.py
```diff
@@ -122,13 +64,62 @@ class DynamicApiViewSet(
...
 
+    def initial(self, request, *args, **kwargs):
+        super().initial(request, *args, **kwargs)
+        self.temporal = TemporalTableQuery(request, self.model.table_schema())
+
+    @cached_property
+    def lookup_field(self):
+        """Overwritten from GenericAPIView to filter on temporal identifier instead."""
+        if self.temporal.is_versioned:
+            return self.model.table_schema().identifier[0]
```
gkoller, line 83: Same here. Why only the first element of `identifier`?

See also https://github.com/Amsterdam/dso-api/pull/389/files#r742560983

src/dso_api/dynamic_api/views/api.py
```diff
@@ -122,13 +64,62 @@ class DynamicApiViewSet(
...
+        else:
+            return "pk"
+
+    def get_queryset(self):
```
gkoller, line 87: `-> models.QuerySet` ?

src/dso_api/dynamic_api/views/api.py
```diff
@@ -122,13 +64,62 @@ class DynamicApiViewSet(
...
+        queryset = super().get_queryset()
+        # Apply the ?geldigOp=... filters, unless ?volgnummer=.. is requested.
+        # The version_value is checked for here, as the TemporalTableQuery can be
+        # checked for a sub object too.
+        if self.temporal and not self.temporal.version_value:
+            queryset = self.temporal.filter_queryset(queryset)
+        return queryset
+
+    def get_object(self):
```
gkoller, line 96: `-> models.Model`? (not sure)

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+        if not self.is_versioned or self.slice_value == "*":
+            # allow this method to be called unconditionally on objects,
+            # and allow ?geldigOp=* to return all data
+            return queryset
+
+        # Either take a given ?geldigOp=yyyy-mm-dd OR ?geldigOp=NOW()
+        slice_value = self.slice_value or now()
+        range_fields = self.slice_range_fields or self.default_range_fields
+
+        if range_fields is not None:
```
gkoller, line 108: More idiomatic: `if range_fields:`

src/dso_api/dynamic_api/serializers.py
```diff
@@ -130,18 +100,20 @@ class DynamicListSerializer(DSOModelListSerializer):
...
             if cast(DynamicModel, related_model).is_temporal() and is_many:
 
-                # When the model is a temporal relationship, make sure the prefetch only
-                # returns the latest objects. The Prefetch objects allow adding these filters.
+                # When the model is a temporal relationship, make sure the prefetch only returns
+                # the proper temporal slice. The Prefetch objects allow adding these filters.
                 lookups[i] = models.Prefetch(
-                    lookup, queryset=filter_latest_temporal(related_model.objects)
+                    lookup,
+                    queryset=filter_temporal_slice(request, related_model.objects.all()),
```
gkoller, line 116: You already know you are dealing with a temporal model (see line 110). Why should `filter_temporal_slice` check that again? Maybe you should simply inline the call to `TemporalTableQuery`

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.slice_dimension = None
+        self.slice_value = None
+        self.slice_range_fields = None
+
+        if temporal is not None:
+            # See if a filter is made on a specific version
+            self.version_field = temporal.identifier  # e.g. "volgnummer"
+            if version := request.GET.get(self.version_field):
+                self.version_value = version
```
gkoller, line 57: No if-statement and intermediate variable necessary
 ```python
self.version_value = request.GET.get(self.version_field, False)
```

File: src/dso_api/dynamic_api/temporal.py [hunk 14]
```diff
@@ -0,0 +1,144 @@
...
+
+        # Parsed arguments are None unless there is a match with the request
+        self.version_field = None
+        self.version_value = None
+
+        self.slice_dimension = None
+        self.slice_value = None
+        self.slice_range_fields = None
+
+        if temporal is not None:
```
gkoller, line 53: `if temporal:`

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+                    self.slice_range_fields = fields
+
+    def _parse_date(self, dimension: str, value: str) -> Union[Literal["*"], date, datetime]:
+        """Parse and validate the received date"""
+        if value == "*":
+            return "*"
+
+        try:
+            if "T" in value or " " in value:
+                return make_aware(datetime.fromisoformat(value))
```
gkoller, line 73: if there is a "T" or an " " in the ISO formatted string that `datetime.fromisoformat` already returns an aware datetime object. So why the `make_aware`?

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+            else:
+                return date.fromisoformat(value)
```
gkoller, line 75: Is the distinction between `date` and `datetime` important? If not  `datetime.fromisoformat` works equally well on date-only strings. You do get back a `datetime` object in that case.

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+        if queryset.model.table_schema() != self.table_schema:
+            raise ValueError("QuerySet model type does not match")
+
+        if not self.is_versioned or self.slice_value == "*":
+            # allow this method to be called unconditionally on objects,
+            # and allow ?geldigOp=* to return all data
+            return queryset
+
+        # Either take a given ?geldigOp=yyyy-mm-dd OR ?geldigOp=NOW()
+        slice_value = self.slice_value or now()
```
gkoller, line 105: Elsewhere you are making sure you are dealing with aware `datetime` objects. But here you are using a naive `datetime` object (as returned by `now()`. Why does it not matter here?

src/dso_api/dynamic_api/temporal.py
```diff
@@ -0,0 +1,144 @@
...
+            return queryset.distinct(identifier).order_by(identifier, f"-{sequence_name}")
+
+    @property
+    def default_range_fields(self) -> Optional[TemporalDimensionFields]:
+        """Tell what the default fields would be to filter temporal objects."""
+        dimensions = self.table_schema.temporal.dimensions
+        return dimensions.get("geldigOp") or first(dimensions.values(), None)
+
+    @property
+    def url_parameters(self) -> dict[str, str]:
```
gkoller, line 130: I don't think this return type is correct. You might be returning a `self.slice_value` as the value in a dict. That instance variable is elsewhere defined to be: `slice_value: Optional[Union[Literal["*"], date, datetime]]. Hence the return type of this function should at least be: `-> dict[str, Union[str, date, datetime]]`

I am kind of amazed that `mypy` didn't caught this. Did you run it?
`

src/dso_api/dynamic_api/serializers.py
vdboor: This is what the old code also does. If there is a compound key `["identifier", "volgnummer"]`, only the grouping first key will be used here.

src/dso_api/dynamic_api/temporal.py
vdboor: I'm using `django.utils.timezone.now()` which is aware.

src/dso_api/dynamic_api/temporal.py
vdboor: It doesn't look like it does:

```
>>> from django.utils.timezone import is_aware
>>> from datetime import datetime
>>> is_aware(datetime.fromisoformat("2021-11-09 12:15:00"))
False
```

src/dso_api/dynamic_api/temporal.py
vdboor: I rather don't check `Temporal.__bool__()` here (had that there before), and only need to know whether it's None.

src/dso_api/dynamic_api/temporal.py
vdboor: good catch!

src/dso_api/dynamic_api/serializers.py
vdboor: The `filter_latest_temporal()` is called from many other places, and doesn't require some pre-assert check before being called. In this situation it's a double check, but on other code paths it makes things much simpler.

src/dso_api/dynamic_api/temporal.py
vdboor: I only want to know whether it's initialized, `is None` does that for me (and is more efficient too).

src/dso_api/dynamic_api/views/api.py
vdboor: :-)

src/dso_api/dynamic_api/views/api.py
vdboor: 😃

src/dso_api/dynamic_api/views/api.py
vdboor: It's the same behavior from the old code. I've added a comment about it:

`# Take the first field as grouper from ["identificatie", "volgnummer"]`

src/dso_api/dynamic_api/temporal.py
vdboor: good point!

src/dso_api/dynamic_api/temporal.py
vdboor: good point, thanks!

src/dso_api/dynamic_api/serializers.py
vdboor: I've added a comment:

`# When 'identifier' is ["identificatie", "volgnummer"], take the first as grouper.`

src/dso_api/dynamic_api/temporal.py
vdboor: yes, I don't want the SQL to make comparisons against a 00:00:00 time if all we have is a date. So `datetime > date` becomes tomorrow, not everything past midnight.

src/dso_api/dynamic_api/temporal.py
vdboor: Good catch! It should format the self.slice_value back to a string.

src/dso_api/dynamic_api/temporal.py
gkoller: Ah, you are right. I thought `isoformat` always included the timezone, hence `fromisoformat` would be timezone aware. But that's not the case. Learned something new. :+1:

PR #391, opened 2021-11-04
Putting DSO-API on a diet: remove django.contrib.auth/contenttypes/messages

vdboor:
These apps are not used at all, as the project doesn't need User models.
Nor do their tables have to be created.


lbam (approved):

lbam:
Build failure appears unrelated:

```
Running style checks
./dso_api/dynamic_api/serializers.py:792:1: C901 '_through_serializer_factory' is too complex (10)
1     C901 '_through_serializer_factory' is too complex (10)
```

PR #392, opened 2021-11-04
AB#11748 Dotted filter syntax on subfields

DonQueso89:

* filterset_factory

NOTE that the openapi docs are still not correctly picking up the filter. I will do that in a follow-up PR.

### DIFF
--- a/src/dso_api/dynamic_api/filterset.py
+++ b/src/dso_api/dynamic_api/filterset.py
@@ -7,12 +7,15 @@ This file provides the translation and required bits for dynamic models.
 """
 import logging
 import re
+from contextlib import suppress
 
 from django.contrib.gis.db.models.fields import MultiPolygonField, PolygonField
 from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import FieldDoesNotExist
 from django.db import models
+from django.db.models.fields.related import ForeignKey
 from django_postgres_unlimited_varchar import UnlimitedCharField
-from schematools.contrib.django.models import DynamicModel
+from schematools.contrib.django.models import DynamicModel, get_field_schema
 from schematools.utils import to_snake_case, toCamelCase
 
 from rest_framework_dso import filters as dso_filters
@@ -32,6 +35,7 @@ DEFAULT_LOOKUPS_BY_TYPE = {
     models.CharField: _string_lookups,
     UnlimitedCharField: _string_lookups,
     models.IntegerField: _comparison_lookups + ["in"],
+    models.BigIntegerField: _comparison_lookups + ["in"],
     models.FloatField: _comparison_lookups + ["in"],
     models.DecimalField: _comparison_lookups + ["in"],
     models.DateField: _comparison_lookups,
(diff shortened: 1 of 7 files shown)


src/dso_api/dynamic_api/filterset.py
```diff
@@ -72,19 +75,60 @@ def filter_to_camel_case(value):
...
+        if isinstance(f, ForeignKey):
+            fields[f.attname] = _get_field_lookups(f)  # backwards compat
+
+            # In case of a composite FK, get the loose relations
+            # associated with this ForeignKey object and add dotted syntax.
+            schema_field = get_field_schema(f)
+            prefix = f.attname.removesuffix("_id")
+            try:
+                for s in schema_field.sub_fields:
+                    db_field_name = to_snake_case(s.id)
```
DonQueso89, line 101: is this guranteed to be the name of the foreign key?

src/dso_api/urls.py
```diff
@@ -15,7 +16,8 @@
         views.multiple_slashes,
         name="error-trailing-slashes",
     ),
-]
+] + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```
DonQueso89, line 19: added this because docker uses uwsgi, in which case the Django DEBUG magic doesnt apply

src/dso_api/dynamic_api/filterset.py
```diff
@@ -72,19 +75,60 @@ def filter_to_camel_case(value):
...
+                        filters[sub_filter_name] = filter_class(
+                            field_name=model_field_name,
+                            lookup_expr=lookup,
+                            label=dso_filters.DSOFilterSet.FILTER_HELP_TEXT.get(
+                                filter_class, lookup
+                            ),
+                        )
+            except ValueError:
+                # The schema field has no sub fields
+                pass
```
gkoller, line 128: In cases where the exception is silenced (eg a single `pass` statement) I prefer the following syntax:

```python
from contextlib import suppres
...
with suppress(ValueError):
   ...
```
This makes it clear, right from the start that we are going to ignore `ValueErrors`. Whereas with a `try` statement the expection is that we are going to catch it and doing something with it, only to read after 16 lines of code we are going to ignore it afterall.

gkoller: Anyway it is just a preference. It's okay to leave it as-is. Just wrote it as not many people are aware of `suppress`' existence.

src/tests/test_dynamic_api/test_filters.py
```diff
@@ -436,3 +436,74 @@ def test_filter_isempty(parkeervakken_parkeervak_model, filled_router):
...
+        }
+
+        assert expected_fields.issubset(
+            [(k, v.field_name) for k, v in filterset_class.get_filters().items()]
+        )
+
+        assert VerblijfsObjecten.objects.count() == 4
+
+        filterset = filterset_class({"heeftHoofdadres.volgnummer": 1})
+        assert filterset.is_valid(), filterset.errors
```
gkoller, line 482: By supplying a message (eg `filterset.errors`) to the `assert` statement, you loose the assertion introspection that pytest gives you. Was that your intention?

See also: https://docs.pytest.org/en/6.2.x/assert.html

src/dso_api/settings.py
```diff
@@ -405,3 +405,9 @@
 HAAL_CENTRAAL_API_KEY = os.getenv("HAAL_CENTRAAL_API_KEY", "UNKNOWN")
 HAAL_CENTRAAL_KEYFILE = os.getenv("HC_KEYFILE")
 HAAL_CENTRAAL_CERTFILE = os.getenv("HC_CERTFILE")
+
+SHELL_PLUS_POST_IMPORTS = [
+    "from django.apps.registry import apps",
+    "from dso_api.dynamic_api.filterset import filterset_factory",
+    "from pprint import pprint",
+]
```
gkoller, line 413: This could be regarded as nitpicking, but I am a firm believer in that code should express intention. Here I see a constant (because the variable is in all caps), but with a mutable data structure (a list) assigned to it. Using a mutable data structure instead of an immutable one suggests, at least to me, that mutability is required. But that contradicts its assignment to a constant.

So I would either suggest, not using a constant here (`SHELL_PLUS_POST_IMPORTS` -> `shell_plus_post_imports`) or use an immutable datastructure (eg a tuple) instead of a list.

gkoller: The same reasoning can be applied to https://github.com/Amsterdam/dso-api/pull/392/files#diff-6b63d8df4a47199509b132ee48d6a28acb9e7daf592fe5bfbb4c20dcd03d79d9R116.

```python
if lookup not in ["contains", "exact"]:
    ...
```
Testing for membership does not require a mutable datastructure. I would write that as:

```python
if lookup not in ("contains", "exact"):
    ...
```
But this is even more nitpicking than then the constant versus list assigment I made above.

src/tests/test_dynamic_api/test_filters.py
```diff
@@ -436,3 +436,74 @@ def test_filter_isempty(parkeervakken_parkeervak_model, filled_router):
         data = read_response_json(response)
         assert len(data["_embedded"]["parkeervakken"]) == 1
         assert data["_embedded"]["parkeervakken"][0]["id"] == "121138489006"
+
+    @staticmethod
```
gkoller, line 440: I do realize that you are merely following the existing code @DonQueso89 , but why are these tests in a class and is each test function annotated with `@staticmethod`? Pinging @vdboor  

 We are using `pytest` here. Hence we can do away with the `unittest`-isms.

I say ditch the class and the `@staticmethod` annotations.

gkoller (approved):

src/dso_api/dynamic_api/filterset.py
DonQueso89: nice, I did not know about this

src/tests/test_dynamic_api/test_filters.py
DonQueso89: I think the message, in this case, allows for quicker insight into what is the root of the test failure, since the introspection would provide `assert False == True where False = is_valid()`

src/dso_api/settings.py
DonQueso89: done :)

src/tests/test_dynamic_api/test_filters.py
DonQueso89: I leave this for discussion

PR #393, opened 2021-11-05
AB#22547 Swagger UI KeyCloak authentication fixed

Yrob:
KeyCloak Authentication in Swagger was broken.
The authorization uri was outdated.
It was also requesting access with limited scopes, which was not supported
by the KeyCloak server.

Also improved the environment variables so they can be used on
the different platforms without custom logic.


vdboor (approved):
cool that this is fixed. I would love to see more documentation on this. For example, that `OAUTH_DEFAULT_SCOPE` avoids needing exceptions for Azure-vs-Keycloak.

Some of the details you've shared on Slack could be introduced in the documentation.

PR #394, opened 2021-11-08
Fix output format of HC-BRK remote

lbam:
The part that belongs inside `_embedded.kadastraalonroerendezaken` was the entire output and the `_links` parts were missing, except for the `schema`, which was in the wrong place.

### DIFF
--- a/src/dso_api/dynamic_api/remote/views.py
+++ b/src/dso_api/dynamic_api/remote/views.py
@@ -2,8 +2,10 @@
 Currently it mainly performs authorization, data retrieval, and schema validation.
 """
 import logging
+from typing import Optional
 
 from django.core.exceptions import ImproperlyConfigured
+from more_ds.network.url import URL
 from rest_framework import status
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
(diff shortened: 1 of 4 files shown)


src/dso_api/dynamic_api/remote/views.py
```diff
@@ -90,7 +92,41 @@ def list(self, request, *args, **kwargs):
         # TODO: add pagination:
         # paginator = self.pagination_class()
         # paginator.get_paginated_response(serializer.data)
-        return Response(serializer.data)
+
+        # Work around the serializer producing the wrong format.
+        # TODO: fix the serializer instead? What we get from the Kadaster remote
+        #  looks more what we want than what comes out of the serializer.
```
vdboor, line 98: indeed. Let's fix the serializer instead.

lbam: Not in this PR, though. I was working on a different issue and just need to get this out of the way.

lbam: Any other comments? I'd like to get on with what I was actually doing.

vdboor (approved):
OK, in overleg approved for now. :-) Kunnen we later beter maken (aangezien  retrieve() al hetzelfde doet)

PR #397, opened 2021-11-10
Fix the heisenbug that happens during runserver initialization

vdboor:
When anything inside the router construction failed, this exception would be silenced. However, it did leave incomplete model state behind that would cause problems further down the line with model relations.

Typically, a ManyToManyField would be unable to resolve their reverse-names, causing the unrelated error:
"'ManyToManyField' object has no attribute '_m2m_reverse_name_cache'".

On inspection, both models has an identical 'CREATION_COUNTER' but a different id causing any `field.related.model == model` tests to fail. The Django apps still had references to the older models that were created during the main runserver process.


jjmurre (dismissed):
Very subtle bug!

lbam (approved):

PR #401, opened 2021-11-15
Remove spaces in BAG schema field ids

lbam:
#399 removed spaces from fields in test schemas. This removes spaces from the copy of the BAG schema. We're not supposed to support those.

This should be a purely cosmetic change.


gkoller (approved):

lbam:
Jenkins failure is the familiar `Network test_default` stuff. I thought we'd fixed that. Merging anyway as the fault is not related.

PR #402, opened 2021-11-15
Don't suppress all ValueErrors in filterset_factory

lbam:
The previous code suppressed all ValueErrors instead of just the specific one indicating "no subfields on this field".

As a result, invalid filters were silently ignored, leading to `[gt]`, `[lt]`, `[gte]` and `[lte]` filters not being executed.

### DIFF
--- a/src/dso_api/dynamic_api/filterset.py
+++ b/src/dso_api/dynamic_api/filterset.py
@@ -97,33 +97,39 @@ def filterset_factory(model: type[DynamicModel]) -> type[DynamicFilterSet]:  # n
             # associated with this ForeignKey object and add dotted syntax.
             schema_field = get_field_schema(f)
             prefix = f.attname.removesuffix("_id")
-            with suppress(ValueError):  # The schema field has no sub fields
-                for s in schema_field.sub_fields:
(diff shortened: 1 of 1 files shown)


src/dso_api/dynamic_api/filterset.py
```diff
@@ -97,33 +97,39 @@ def filterset_factory(model: type[DynamicModel]) -> type[DynamicFilterSet]:  # n
...
-                            label=dso_filters.DSOFilterSet.FILTER_HELP_TEXT.get(
-                                filter_class, lookup
-                            ),
-                        )
+
+            try:
+                subfields = schema_field.sub_fields
+            except ValueError as e:
+                # Check for the specific ValueError that signals no subfields.
+                if "subfields are only possible" in str(e).lower():
```
jjmurre, line 105: It's a pity we do not get more specific exceptions here, but this comes straight from a library right?

lbam: This comes from schema-tools. I'm working on a patch for that, but I suggest merging this first.

jjmurre (approved):
OK

PR #411, opened 2021-11-18
Properly resolve database source fields in temporal _links

DonQueso89:
Fixes [AB#24665](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/24665)

The temporal _links factory was directly using the camel-cased values from the schema identifier field to add serializer fields by name. This did not lead to an error as long as the fields in the identifier of temporal schemas were equal to the camel-cased name used as the eventual database field name (i.e.: within ams schema context: they did not use shortnames and did not use uppercase letters) . These are now properly resolved by using the snakecased `DatasetFieldSchema.name` as the source of the field.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -579,6 +579,9 @@ def serializer_factory(
         nesting_level,
     )
 
+    # TODO: Resolve the actual names of the fields in the database
+    # by doing `table_schema.get_field_id(x).name for x in identifier`
+    # then do the same when matching them below with the model_field name
     if table_schema.temporal:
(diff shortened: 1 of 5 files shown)


src/dso_api/dynamic_api/serializers.py
```diff
@@ -597,10 +600,10 @@ def serializer_factory(
         if (
             # Do not render PK and FK to parent on nested tables
             (model.has_parent_table() and model_field.name in ["id", "parent"])
```
jjmurre, line 602: Looks good! Would it be possible to add a test for this?

PR #413, opened 2021-11-22
Allow ?_field=... to work across relations too.

vdboor:
This required a completely different parser (FieldsToDisplay) that can handle the various include/exclude patterns across different nesting levels. The final result still produces an allowlist of all fields that should be preserved, and which embeddings to preserve.

### DIFF
--- a/src/rest_framework_dso/fields.py
+++ b/src/rest_framework_dso/fields.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Optional, Union
+from typing import Iterable, Optional, Union
 from urllib.parse import quote, urlsplit, urlunsplit
 
 from django.contrib.gis.geos import GEOSGeometry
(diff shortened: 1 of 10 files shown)


jjmurre (dismissed):
LGTM, Nice refactor! Especially the tests are a great help to understand what is going on. These kind of examples/scenario's really should go into the user docs.

src/rest_framework_dso/fields.py
```diff
@@ -9,21 +9,234 @@
...
 
-def parse_request_fields(fields: Optional[str]):
-    if not fields:
-        return None
 
-    # TODO: support nesting, perform validation
-    return fields.split(",")
+ALLOW_ALL_FIELDS_TO_DISPLAY = FieldsToDisplay()
+DENY_SUB_FIELDS_TO_DISPLAY = FieldsToDisplay()
+DENY_SUB_FIELDS_TO_DISPLAY._allow_all = False
```
DonQueso89, line 239: is it safe to re-use these instances when determining nested representations for different sub fields?

DonQueso89 (approved):

src/rest_framework_dso/fields.py
vdboor: good point, yes for now. The objects are not mutated themselves after init. (unless you start meddling with the returned sets perhaps)

vdboor:
hmm force-pushing a comment fix invalidated all approvals. Hence have merged this using admin rights.

src/rest_framework_dso/fields.py
gkoller: The are marked up as contstants, hence should be type annotated as such:

```python
ALLOW_ALL_FIELDS_TO_DISPLAY: Final = FieldsToDisplay()
DENY_SUB_FIELDS_TO_DISPLAY: Final = FieldsToDisplay()._allow_all = False
```

PR #415, opened 2021-11-25
AB#24678 Fix test_views_api.py

lbam:
The auth tests assumed they would get 200 when filtering on a field that doesn't exist. That conflicts with the requirement of getting a 400 in this case. Add the field (`"volgnummer"`) and check for 404 instead, since we can't deliver the actual 400 yet.

There's a test that requires the same behavior for `"regimes.inWerkingOp"`, which was removed for [AB#24402](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/24402). This still needs updating.

### DIFF
--- a/src/tests/files/parkeervakken.json
+++ b/src/tests/files/parkeervakken.json
@@ -21,6 +21,10 @@
           "schema": {
             "$ref": "https://schemas.data.amsterdam.nl/schema@v1.1.0#/definitions/schema"
           },
+          "volgnummer": {
+            "type": "integer",
+            "description": ""
+          },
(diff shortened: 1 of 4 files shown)


vdboor (dismissed):

vdboor:
prima! Die test zou ook inderdaad beter mogen :-) Iets meer checken op de inhoud van het resultaat

lbam:
Forgot to run and update the other tests that use the `parkeervakken` dataset. Re-review?

PR #355, opened 2021-09-22
Azure dbs

ngaranko:
# This Pull request contains changes to:

Ability to get data from KeyVault using CSI provider on Azure AKS.

# In case changes new features added

- [ ] Customer friendly documentation added into `docs/`.
- [x] Developer friendly documentation added into `DEVELOPMENT|README.md`

PR #418, opened 2021-11-29
Retrieve the PGPASSWORD from env var, not from secrets mount

tlrh314:


src/dso_api/settings.py
```diff
@@ -142,20 +142,17 @@
...
-                "HOST": env.str("PGHOST"),
-                "PORT": env.str("PGPORT"),
-                "OPTIONS": {"sslmode": "require"},
-            }
+    DATABASES = {
+        "default": {
+            "ENGINE": "django.contrib.gis.db.backends.postgis",
+            "NAME": env.str("PGDATABASE"),
+            "USER": env.str("PGUSER"),
+            "PASSWORD": env.str("PGPASSWORD"),
```
jjmurre, line 150: Could you also describe the "why" of the PR in the commit message.

src/dso_api/settings.py
```diff
@@ -142,20 +142,17 @@
 }
 
 if CLOUD_ENV.lower().startswith("azure"):
-    with open("/mnt/secrets-store/mdbdataservices-read") as secrets_file:
```
jjmurre, line 145: Is it possible to trace back from the commit messages why this was done this way?

tlrh314: I've updated the commit message to be more verbose on the why.

src/dso_api/settings.py
tlrh314: I've updated the commit message to be more verbose on the why.

src/dso_api/settings.py
tlrh314: The current implementation comes from https://github.com/Amsterdam/dso-api/pull/355, which probably overlooked that the env var `PGPASSWORD` is exported in the k8s manifest in addition to volume mounting it.

lbam (approved):

PR #419, opened 2021-11-29
Improve error message when coordinate conversion fails during rendering

vdboor:
This directs users to the data supplier instead of having to ping us for support. Any exception that inherits from HumanReadableException is shown in full detail to end users on errors.

### DIFF
--- a/src/rest_framework_dso/exceptions.py
+++ b/src/rest_framework_dso/exceptions.py
@@ -42,3 +42,17 @@ class RemoteAPIException(APIException):
         self.code = code or self.default_code
         self.default_detail = title
         self.status_code = status_code
+
+
+class HumanReadableException(APIException):
+    """Base class for tagging exceptions that can be shown in complete detail to the client.
+    This can be shown in the standard exception handler, or during streaming rendering.
+    """
+
+    default_detail = "Error during rendering"
+
+
+class HumanReadableGDALException(HumanReadableException):
+    """Tell that rendering failed due to conversion errors"""
+
+    default_detail = "Coordinate conversion failed"
--- a/src/rest_framework_dso/serializers.py
+++ b/src/rest_framework_dso/serializers.py
@@ -22,6 +22,7 @@ import inspect
 from typing import Iterable, Optional, Union, cast
 
 from django.contrib.gis.db import models as gis_models
+from django.contrib.gis.gdal import GDALException
 from django.db import models
 from django.utils.functional import cached_property
 from rest_framework import serializers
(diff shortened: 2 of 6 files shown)


src/rest_framework_dso/serializers.py
```diff
@@ -569,7 +571,19 @@ def _apply_crs(self, instance, accept_crs: CRS):
...
-                accept_crs.apply_to(geo_value)
+                try:
+                    accept_crs.apply_to(geo_value)
+                except GDALException as e:
+                    # While there could be various reasons for this, the most common one
+                    # is that the data has coordinates outside the projected bounds of the CRS.
+                    # Instead rendering /* Aborted by GDALException during rendering! */ this
+                    # exception message points the user directly to the data supplier.
+                    raise HumanReadableGDALException(
+                        "Fout tijdens coördinaat conversie voor"
```
lbam, line 582:
```suggestion
                        "Fout tijdens coördinaatconversie voor"
```

lbam:
Test?

Also, I'm curious where this error pops up, in the middle of the stream as before or instead of the entire stream?

src/rest_framework_dso/serializers.py
vdboor: tx!

vdboor:
> Also, I'm curious where this error pops up, in the middle of the stream as before or instead of the entire stream?

depends on the record. As the first record is parsed *before* the streaming starts, this triggers the DRF exception handler. Any other record will cause an error during the rendering. I've noticed both scenario's could happen and these are both addressed in this PR.

src/rest_framework_dso/exceptions.py
```diff
@@ -42,3 +42,17 @@ def __init__(self, title, detail, code, status_code):
...
         self.default_detail = title
         self.status_code = status_code
+
+
+class HumanReadableException(APIException):
+    """Base class for tagging exceptions that can be shown in complete detail to the client.
+    This can be shown in the standard exception handler, or during streaming rendering.
+    """
+
+    default_detail = "Error During Rendering"
```
lbam, line 52:
```suggestion
    default_detail = "error during rendering"
```

src/rest_framework_dso/exceptions.py
```diff
@@ -42,3 +42,17 @@ def __init__(self, title, detail, code, status_code):
...
+
+
+class HumanReadableGDALException(HumanReadableException):
+    """Tell that rendering failed due to conversion errors"""
+
+    default_detail = "Coordinate Conversion Failed"
```
lbam, line 58:
```suggestion
    default_detail = "coordinate conversion failed"
```

src/rest_framework_dso/exceptions.py
vdboor: I've used uppercase-first instead of title-case, which is indeed in line with other APIException classes from DRF.

lbam (approved):

PR #421, opened 2021-11-29
Fix test schema's and code that uses DatasetSchema internal representation.

gkoller:
New version of `schema-tools` keeps track of active table versions. This is entirely transparant to all code that only uses `DatasetSchema`'s public API (methods and properties). Code that uses the underlying `dict` representation of `DatasetSchema`, however, will be affected. This PR has the necessary changes for those pieces of code.


jjmurre (approved):
LGTM

PR #423, opened 2021-11-30
AB#24404 Whitelist "page_size" in filter auth

lbam:
I forgot to whitelist this parameter. Of the tests, the one in `test_rest_framework_dso` doesn't actually trigger the bug, but I included it for completeness. The one in `test_dynamic_api` does, but it triggers a different bug, [AB#24678](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/24678).


jjmurre (approved):
LGTM

PR #424, opened 2021-11-30
Do not include pk of non temporal targets when serializing through tables

DonQueso89:
I did this in a previous PR and thought it was covered by tests but that turned out to be a mistake. This requires a closer look at how through serialization works because now it is breaking the serialization of m2m relations. I will do this in a follow-up PR.

### DIFF
--- a/src/dso_api/dynamic_api/serializers.py
+++ b/src/dso_api/dynamic_api/serializers.py
@@ -978,15 +978,13 @@ def _through_serializer_factory(  # noqa: C901
             "title", serializers.CharField(source=title_field, read_only=True)
         )
 
-    # Always include the primary identifier of the targeted relationship,
-    id_seq = first(target_table_schema.identifier)  # e.g.: "identificatie"
-    serializer_part.add_field_name(id_seq, source=f"{target_fk_name}_{id_seq}")
-
     # See if the table has historical data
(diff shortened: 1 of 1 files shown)


src/dso_api/dynamic_api/serializers.py
```diff
@@ -978,15 +978,13 @@ def _through_serializer_factory(  # noqa: C901
...
         )
 
-    # Always include the primary identifier of the targeted relationship,
-    id_seq = first(target_table_schema.identifier)  # e.g.: "identificatie"
-    serializer_part.add_field_name(id_seq, source=f"{target_fk_name}_{id_seq}")
-
     # See if the table has historical data
     temporal: Optional[Temporal] = target_table_schema.temporal
     if temporal is not None:
         # Include the temporal identifier of the targeted relationship,
```
Yrob, line 984: Moet de commentaar regel niet ook mee verhuisd worden?

Yrob (approved):

src/dso_api/dynamic_api/serializers.py
DonQueso89: nope, die was alleen van toepassing op de oude code. Nu voegen we de pk alleen toe als de target temporal is

PR #425, opened 2021-11-30
AB#24579 fix to_snake_case bug in router

Yrob:
_build_db_viewsets was registering views without calling to_snake_case
on the dataset id. This caused beheerkaart_basis and beheerkaart_cbs grid to give a 500.


lbam:
Needs a unit test, but approved anyway.

PR #426, opened 2021-11-30
Hotfix for filter authorization on BBGA

lbam:


jjmurre (approved):
LGTM Let's get it out of the way first.

PR #431, opened 2021-12-01
Document pagination parameter "page"

lbam:
This wasn't explained yet, and we got confused about it ourselves this morning.


jjmurre (approved):
LGTM, explicit is better.

PR #436, opened 2021-12-08
AB#24404 AB#24678 Filter validation

lbam:
Filter validation+authorization now happens in DynamicApiViewSet.initial, much earlier in the call chain.

The check still lives in permissions.py, which may not be right place for it.

### DIFF
--- a/src/dso_api/dynamic_api/permissions.py
+++ b/src/dso_api/dynamic_api/permissions.py
@@ -86,10 +86,6 @@ class HasOAuth2Scopes(permissions.BasePermission):
 
         schema = model.table_schema()
         access = request.user_scopes.has_table_access(schema)
-        if access and not _filters_ok(request, schema):
-            # Don't give access when the user is filtering on fields they may not see.
-            access = Permission.none
-
         log_access(request, access)
         return access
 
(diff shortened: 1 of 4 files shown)


src/dso_api/dynamic_api/permissions.py
```diff
@@ -142,35 +149,33 @@ def _filters_ok(request, schema: DatasetTableSchema) -> bool:  # noqa: C901
...
 
     for key, values in request.GET.lists():
-        if key in ("fields", "format", "page", "page_size", "sorteer") or key.startswith("_"):
+        if key in allowed:
             continue
 
         # Everything else is a filter.
         field_name, op = _parse_filter(key)
 
+        # OPTIONS requests should validate, but are not
```
jjmurre, line 158: Seems like the sentence is not finished?

DonQueso89 (approved):

src/dso_api/dynamic_api/permissions.py
lbam: This shouldn't have been here at all. OPTIONS is already checked at function entry time. Removed.

jjmurre (approved):
LGTM. Code is clean and readable. I cannot see all implications, but the tests are helpfull in understanding what's going on.

PR #437, opened 2021-12-09
AB#24404 Handle syntax errors in filter parameters

lbam:
These would produce 500 instead of 400 due to a NameError.

It may seem overkill to introduce an exception class for this, but

* ValueError is too broad: it can be caused by a bug as well as a
  validation check.
* Django's ValidationError isn't handled by DRF, so it would require
  changes to the exception handler. I'm not willing to go there right
  now.
* DRF's ValidationError may cause problems with non-DRF views such as
  remotes. We want reusable validation/authorization code.

> Don't forget about...
> * Tests
> * Documentation in `docs/`, `dev-docs/`
>

### DIFF
--- a/src/dso_api/dynamic_api/permissions.py
+++ b/src/dso_api/dynamic_api/permissions.py
@@ -122,7 +122,7 @@ def validate_request(request, schema: DatasetTableSchema, allowed: set[str]) -> 
     The argument allowed is a set of parameters that are explicitly allowed.
     All other parameters are parsed as filters and rejected if not present in the schema.
 
-    Raises SchemaObjectNotFound or PermissionDenied in case of a problem.
+    Raises FilterSyntaxError, SchemaObjectNotFound or PermissionDenied in case of a problem.
     """
 
     if request.method == "OPTIONS":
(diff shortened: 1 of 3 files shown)


src/dso_api/dynamic_api/permissions.py
```diff
@@ -224,7 +230,7 @@ def _parse_filter(v: str) -> tuple[str, str]:
         field_name = v
         op = "exact"
     elif not v.endswith("]"):
-        raise ValidationError(f"missing closing bracket (]) in {v!r}")
+        raise FilterSyntaxError(f"missing closing bracket (]) in {v!r}")
```
jjmurre, line 233: What happens when an opening bracket is missing. That is alread caught I suppose?

lbam: That produces a 400 with a `not found in schema` message. I can add a case for it to deal with it more cleanly.

jjmurre: Whatever you wish. To me this is already OK.

jjmurre (approved):
Nice catch, LGTM.

src/dso_api/dynamic_api/permissions.py
lbam: Added a check and tests. Please review again.

jjmurre (approved):
LGTM

PR #441, opened 2021-12-13
Added dump_serializers command for debugging

vdboor:
Just like `manage.py dump_models` we now have `manage.py dump_serializers`.

### DIFF
--- a/src/dso_api/dynamic_api/management/commands/dump_serializers.py
+++ b/src/dso_api/dynamic_api/management/commands/dump_serializers.py
@@ -1,0 +1,266 @@
+import inspect
+import textwrap
+from argparse import ArgumentParser
+from typing import Any, List, Optional, Tuple, Type
+
+from django.apps import AppConfig, apps
+from django.core.management import BaseCommand
+from rest_framework import serializers
+from rest_framework.request import Request
+from rest_framework.test import APIRequestFactory
+from schematools.permissions import UserScopes
+from schematools.types import DatasetTableSchema
+
+from dso_api.dynamic_api.serializers import DynamicSerializer
+from dso_api.dynamic_api.urls import router
+from rest_framework_dso.fields import AbstractEmbeddedField
+from rest_framework_dso.renderers import HALJSONRenderer
+from rest_framework_dso.serializers import DSOSerializer
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/management/commands/dump_serializers.py
```diff
@@ -0,0 +1,266 @@
...
+                doc += "\n"
+        self.stdout.write(f'    """{doc}"""\n\n')
+
+    def write_factory_docs(self, serializer_class: Type[DynamicSerializer]) -> None:
+        """Write debugging information about the factory function."""
+        # Write some factory information
+        factory_name = self._get_serializer_factory_name(serializer_class)
+        if issubclass(serializer_class, DynamicSerializer):
+            self.stdout.write(f"    # Set by {factory_name}():\n")
+            self.stdout.write(f"    # table_schema = {serializer_class.table_schema}\n")
```
gkoller, line 169: Minor nitpicking. You seem to be doing identation manually. That soon starts to become errorprone, do I have 7 or 8 spaces (=level 2), 11 or 12 (=level 3), etc. You could do something like this:

```python
ident = lambda level: " " * 4 * level
...
self.stdout.write(f"{ident(1)}# Set by {factory_name}():\n")
self.stdout.write(f"{ident(1)}# table_schema = {serializer_class.table_schema}\n")
...
... 
self.stdout.write(f"{ident(2)}{name} = {value!r}\n")
```
Looks weird, but it is immediately clear what indentation level you are on.

gkoller (approved):

src/dso_api/dynamic_api/management/commands/dump_serializers.py
lbam: Less verbose:

```python
def indented(level, s):
    self.stdout.write(" " * 4 * level)
    self.stdout.write(s)

gkoller: @lbam That's a much better solution. Thanks!

vdboor: well in this case (and with the dump_models) command, there is only one indentation level. This won't change either so I'd vote for keeping the code as it is

PR #442, opened 2021-12-13
Fixed source==field_name issue for hoofdroutes

vdboor:
DRF doesn't want to have `field_name` to be identical to `source`. It assumes this is written out as literal source code, and warns not to have redundant `source=field_name` kwargs.


jjmurre (approved):
LGTM

PR #444, opened 2021-12-15
Handle broken datasets gracefully on /v1/

lbam:
Previously, any error would give a 500 on /v1/. Any breakage now causes a dataset to be skipped in the listing.


jjmurre (approved):
LGTM

PR #445, opened 2021-12-15
Handle broken datasets gracefully on /v1/, now for real

lbam:
Forgot to commit the final version before push.


jjmurre (approved):
Even better!

PR #457, opened 2022-01-17
Add the `maketoken` script as Django mgm. command

jjmurre:
Script taken from the readthedocs (courtesy lbam).


lbam (approved):

lbam:
LGTM, but shouldn't we update the docs to mention this command too?

lbam (approved):

PR #459, opened 2022-01-19
[AB#27820] Refactor point parsing further

lbam:
This is needed for the filter refactor. Existing tests should suffice.


vdboor (approved):
Prima! Zeker aangezien deze oude code straks wegkan.

PR #464, opened 2022-01-25
Fix accessing /v1/wfs/beheerkaart/basis/ due to camel-case dataset name

vdboor:
The `beheerkaartBasis` dataset id was converted into `beheerkaart_basis` when it was accessed via the `Dataset.name` attribute.  This change also reduces the need for an extra query to get the description.


tlrh314 (approved):

jjmurre (approved):
LGTM

PR #467, opened 2022-01-27
Add a filter for Sentry

jjmurre:
To be able to avoid sending events to sentry for specific datasets (urls).

### DIFF
--- a/dev-docs/source/environment.rst
+++ b/dev-docs/source/environment.rst
@@ -47,6 +47,15 @@ This can be changed using::
     DSO_API_LOG_LEVEL = ...
     DSO_API_AUDIT_LOG_LEVEL = ...
 
+For some endpoints, when error messages are generated, those message
+are containing too much confidential information (like BSN numbers) to end up in the Sentry logs.
+
+So, it is possible to ignore certain endpoints based on the `path` part of the url
+using the environment variable `SENTRY_BLOCKED_PATHS`. This environment variable
+should contain a comma-separated list of strings that will flag the endpoint for
+not logging to Sentry when one of the defined strings in this environment variable
+is part of the url. 
+
 
 Dataset Schema Locations
 ------------------------
--- a/src/dso_api/sentry.py
+++ b/src/dso_api/sentry.py
@@ -1,0 +1,25 @@
+from os.path import normpath
+from typing import Any, Optional
+from urllib.parse import urlparse
+
+from django.conf import settings
+
+# The makers of sentry only have internal types (in `_types`)
+# those are not considered stable atm, so we define
+# our own aliases for now.
+Event = dict[str, Any]
+Hint = dict[str, Any]
+
+
+def before_send(event: Event, hint: Hint) -> Optional[Event]:
+    """Filters events before they are sent to the Sentry server."""
+
+    path = urlparse(event["request"]["url"]).path
+
+    if any(
+        path_fragment in normpath(path)
+        for path_fragment in settings.SENTRY_BLOCKED_PATHS
+        if path_fragment
+    ):
+        return None
+    return event
--- a/src/dso_api/settings.py
+++ b/src/dso_api/settings.py
@@ -3,7 +3,7 @@ import logging
 import os
 import sys
 from pathlib import Path
-from typing import Optional
+from typing import Final, Optional
 
 import environ
 import sentry_sdk
@@ -14,17 +14,20 @@ from opencensus.trace import config_integration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import LoggingIntegration
 
+from dso_api.sentry import before_send
+
 env = environ.Env()
 
 # -- Environment
 
 BASE_DIR = Path(__file__).parents[1]
-DEBUG = env.bool("DJANGO_DEBUG", True)
+DEBUG = env.bool("DJANGO_DEBUG", False)
 
 CLOUD_ENV = env.str("CLOUD_ENV", "default")
 DJANGO_LOG_LEVEL = env.str("DJANGO_LOG_LEVEL", "INFO")
 DSO_API_LOG_LEVEL = env.str("DSO_API_LOG_LEVEL", "INFO")
 DSO_API_AUDIT_LOG_LEVEL = env.str("DSO_API_AUDIT_LOG_LEVEL", "INFO")
+SENTRY_BLOCKED_PATHS: Final[str] = env.list("SENTRY_BLOCKED_PATHS", default=[])
 
 # Paths
 STATIC_URL = "/v1/static/"
@@ -169,6 +172,7 @@ if SENTRY_DSN:
     sentry_sdk.init(
         dsn=SENTRY_DSN,
         environment="dso-api",
+        before_send=before_send,
         integrations=[
             LoggingIntegration(event_level=logging.WARNING),
             DjangoIntegration(),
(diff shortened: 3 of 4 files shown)


dev-docs/source/environment.rst
```diff
@@ -47,6 +47,15 @@ This can be changed using::
     DSO_API_LOG_LEVEL = ...
     DSO_API_AUDIT_LOG_LEVEL = ...
 
+For some endpoints, when error messages are generated, those message
+are containing too much confidential information (like BSN numbers) to end up in the Sentry logs.
+
+So, it is possible to blacklist certain endpoints based on the `path` part of the url
+using the environment variable `SENTRY_BLOCKED_PATHS`. This environment variable
+if defined, should contain a comma-separated list of strings that will flag the endpoint for
```
lbam, line 55:
```suggestion
should contain a comma-separated list of strings that will flag the endpoint for
```

Either this, or add a comma before "if defined" (but less words is more better, right?).

src/dso_api/sentry.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any, Final, Optional
+from urllib.parse import urlparse
+
+import environ
+
+env = environ.Env()
+
+# The makers of sentry only have internal types (in `_types`)
+# those are not considered stable atm, so we define
+# our own aliasses for now.
```
lbam, line 10:
```suggestion
# our own aliases for now.
```

lbam:
I'm not convinced that this works 100%. If `v1/brk` is in the deny list and I request `v1//brk`, will the path get normalized before the check?

jjmurre:
> I'm not convinced that this works 100% of the time. If `v1/brk` is in the deny list and I request `v1//brk`, will the path get normalized before the check? Consider
> 
> ```python
> >>> u = urlparse("https://example.com//foo//bar")
> >>> u.path
> '//foo//bar'
> ```

Good point! For `foo//bar` I get a 404. However, `/v1/foo/../bag/` gets normalized to `/v1/bag`. So, I'll add normalization.

lbam:
The 404 is exactly the error message that shouldn't end up in Sentry, right?

src/dso_api/sentry.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+SENTRY_BLOCKED_PATHS: Final[str] = env.str("SENTRY_BLOCKED_PATHS", "")
+
+
+def before_send(event: Event, hint: Hint) -> Optional[Event]:
+    """Filters events before they are sent to the Sentry server."""
+    blocked_paths = SENTRY_BLOCKED_PATHS.split(",")
+    path = urlparse(event["request"]["url"]).path
+    if any(path_fragment in path for path_fragment in blocked_paths if path_fragment):
```
vdboor, line 22: `if set(blocked_paths) & set(path)` ?
en SENTRY_BLOCKED_PATHS kan je eenmalig (bovenaan de file) al parsen.

src/dso_api/sentry.py
```diff
@@ -0,0 +1,24 @@
...
+Event = dict[str, Any]
+Hint = dict[str, Any]
+
+
+SENTRY_BLOCKED_PATHS: Final[str] = env.str("SENTRY_BLOCKED_PATHS", "")
+
+
+def before_send(event: Event, hint: Hint) -> Optional[Event]:
+    """Filters events before they are sent to the Sentry server."""
+    blocked_paths = SENTRY_BLOCKED_PATHS.split(",")
```
vdboor, line 20: Deze kan je toch eenmalig bovenaan de file parsen? en al ontdoen van lege waarden.

src/dso_api/sentry.py
```diff
@@ -0,0 +1,24 @@
...
+env = environ.Env()
+
+# The makers of sentry only have internal types (in `_types`)
+# those are not considered stable atm, so we define
+# our own aliasses for now.
+Event = dict[str, Any]
+Hint = dict[str, Any]
+
+
+SENTRY_BLOCKED_PATHS: Final[str] = env.str("SENTRY_BLOCKED_PATHS", "")
```
vdboor, line 15: `env.list()` gebruiken en dan wordt het al opgeplitst vanuit comma-separated

vdboor: Deze variabele kan je wel gewoon in de settings file houden. Dan werken alle logica (zoals `@override_settings` ook gewoon).

src/dso_api/sentry.py
vdboor: Hier `settings.SENTRY_BLOCKED_PATHS` gebruiken. Hoef je niet nogmaals de `env.Environ()` te initialiseren.

src/dso_api/sentry.py
jjmurre: Ah, idd, handig!

src/dso_api/sentry.py
jjmurre: Heb ik wel overwogen. Maar, deze `sentry.py` module wordt zelf geimporteerd in `settings.py`, dus circular import.

jjmurre: Met een iets andere opzet en de lazy loading werkt dit nu.

jjmurre:
> > The 404 is exactly the error message that shouldn't end up in Sentry, right?
> 
> Exactly!

I tried explicity filtering out 404 messages. However, by default those messages are not logged to Sentry anyway.
The Sentry event information about the request does not have `status` information, so, it is not possible to do an extra check for a 404 status.

dev-docs/source/environment.rst
```diff
@@ -47,6 +47,15 @@ This can be changed using::
     DSO_API_LOG_LEVEL = ...
     DSO_API_AUDIT_LOG_LEVEL = ...
 
+For some endpoints, when error messages are generated, those message
+are containing too much confidential information (like BSN numbers) to end up in the Sentry logs.
+
+So, it is possible to blacklist certain endpoints based on the `path` part of the url
```
tlrh314, line 53: So, it is possible to ignore certain endpoints based on the `path` part of the url

src/dso_api/settings.py
```diff
@@ -14,17 +14,20 @@
...
 
 BASE_DIR = Path(__file__).parents[1]
-DEBUG = env.bool("DJANGO_DEBUG", True)
+DEBUG = env.bool("DJANGO_DEBUG", False)
 
 CLOUD_ENV = env.str("CLOUD_ENV", "default")
 DJANGO_LOG_LEVEL = env.str("DJANGO_LOG_LEVEL", "INFO")
 DSO_API_LOG_LEVEL = env.str("DSO_API_LOG_LEVEL", "INFO")
 DSO_API_AUDIT_LOG_LEVEL = env.str("DSO_API_AUDIT_LOG_LEVEL", "INFO")
+SENTRY_BLOCKED_PATHS: Final[str] = env.list("SENTRY_BLOCKED_PATHS")
```
tlrh314, line 30: `SENTRY_BLOCKED_PATHS: Final[str] = env.list("SENTRY_BLOCKED_PATHS", default=[])`

tlrh314 (changes requested):

dev-docs/source/environment.rst
jjmurre: Oh yeah, good point. Forbidden word has sneaked into the documentation :-)

src/dso_api/settings.py
jjmurre: Good catch! Turns out that `django-environ` does raise an exception, where I assumed that it would default to a `[]` based on the type.

tlrh314 (approved):

PR #469, opened 2022-02-02
Add simple check to trigger 500 errors

jjmurre:
As a kind of reverse health check.

### DIFF
--- a/src/tests/test_health.py
+++ b/src/tests/test_health.py
@@ -5,3 +5,16 @@ import pytest
 def test_healthchecks(client):
     response = client.get("/status/health/")
     assert response.status_code == 200
+
+
+@pytest.mark.django_db
+def test_runtime_error(client):
+    with pytest.raises(RuntimeError):
+        client.get("/500-test/")
+
+
+@pytest.mark.django_db
+def test_runtime_error_status_code(client):
+    client.raise_request_exception = False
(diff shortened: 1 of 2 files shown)


src/tests/test_health.py
```diff
@@ -5,3 +5,10 @@
 def test_healthchecks(client):
     response = client.get("/status/health/")
     assert response.status_code == 200
+
+
+@pytest.mark.django_db
+def test_runtime_error(client):
+    with pytest.raises(RuntimeError):
+        response = client.get("/500-test/")
```
vdboor, line 13: Of hij raised, of hij assert. Maar niet beide. je kan `api_client.raise_request_exception = False` instellen hiervoor.

jjmurre: Ja, idd, die raise check is voldoende, ik haal de assert weg.

lbam (approved):

lbam:
Maybe we should do something like

```python
if settings.TESTING:
    urlpatterns.append(path("500-test/", _raise_runtime_error))
```

then override that setting in the test? This can be done with test suite classes: https://docs.djangoproject.com/en/4.0/topics/testing/tools/#overriding-settings. Not sure how it would work with test functions.

lbam:
Oh, too late...

PR #479, opened 2022-02-24
[AB#24838, AB#30003] Fix types in parkeervakken test schema

lbam:
This schema breaks the new filters, which look at the types in the schema instead of in the DB.


jjmurre (approved):
Wow, surprisingly this never caused any trouble until now.

PR #482, opened 2022-03-04
Don't default to test JWKS (AB#38587)

lbam:
The test key belongs in src/tests/settings.py, not the main settings.py. It also needs to be set when developing on the command line, or DSO-API fails to start with an AuthzConfigurationError. This is inconvenient, but much safer than the old setup.

### DIFF
--- a/dev-docs/source/howto/install.rst
+++ b/dev-docs/source/howto/install.rst
@@ -103,7 +103,8 @@ To address the error we need to mark that dataset as a proxy to another URL::
 Run the server
 --------------
 
-The server needs a JSON Web Key Set (JWKS) to do authorization.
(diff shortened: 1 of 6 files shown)


dev-docs/source/howto/install.rst
```diff
@@ -103,8 +103,16 @@ To address the error we need to mark that dataset as a proxy to another URL::
 Run the server
 --------------
 
+The server needs a JSON Web Key Set (JWKS) to do authorization.
```
jjmurre, line 106: Does the test token have an expiry? If so, maybe add a warning about that.

lbam: Done.

Yrob (approved):

PR #493, opened 2022-03-16
Fix auth check on composite relation fields

lbam:
Relations with composite keys would error out with 500 because the auth/validation check assumed `DatasetFieldSchema.related_field_ids` would return a single value. That is obviously not true for such fields.


jjmurre (approved):
LGTM, kan het niet 100 % beoordelen, maar zie geen gekke dingen.

PR #494, opened 2022-03-21
Remove broken part of relation auth

lbam:


jjmurre (approved):
LGTM. As we decided, let's move on with this and avoid potentially dangerous relations in schemas with a validator.

PR #497, opened 2022-03-30
Fix relation filter validation, again (AB#40932)

lbam:
This should make afvalwijzer work again, though with the old +"Id" syntax.

Tests have been restored, expanded and commented upon. They have also been updated to no longer use an "auth" field on an identifier, since the schema validation in schema-tools no longer allows that (identifiers show up in URLs anyway, so hiding them otherwise makes little sense).

### DIFF
--- a/src/tests/test_dynamic_api/test_request_validation.py
+++ b/src/tests/test_dynamic_api/test_request_validation.py
@@ -59,9 +59,9 @@ def test_check_filter_simple(field_name: str, scopes: str, exc_type: Optional[ty
         ("base.id", "REFERS REFERS/BASE BASE/TABLE", PermissionDenied),
         # Just mentioning the relation name as the field name isn't enough.
         ("base", "REFERS REFERS/BASE BASE BASE/TABLE", FilterSyntaxError),
-        # The baseId syntax doesn't work with composite keys.
-        ("baseId", "REFERS REFERS/BASE BASE BASE/ID", FilterSyntaxError),
-        ("baseId", "REFERS REFERS/BASE BASE BASE/ID BASE/VOLGNR", FilterSyntaxError),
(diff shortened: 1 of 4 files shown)


vdboor (dismissed):

src/tests/test_dynamic_api/test_request_validation.py
```diff
@@ -46,11 +49,19 @@ def test_check_filter_simple(field_name: str, scopes: str, exc_type: Optional[ty
...
+        ("base.id", "REFERS REFERS/BASE BASE BASE/TABLE", None),
+        ("base.volgnr", "REFERS REFERS/BASE BASE BASE/TABLE", None),
+        # This needs access to both the dataset and the table.
+        ("base.id", "REFERS REFERS/BASE BASE", PermissionDenied),
+        ("base.id", "REFERS REFERS/BASE BASE/TABLE", PermissionDenied),
+        # Just mentioning the relation name as the field name isn't enough.
+        ("base", "REFERS REFERS/BASE BASE BASE/TABLE", FilterSyntaxError),
+        # The baseId syntax doesn't work with composite keys.
+        ("baseId", "REFERS REFERS/BASE BASE BASE/ID", FilterSyntaxError),
+        ("baseId", "REFERS REFERS/BASE BASE BASE/ID BASE/VOLGNR", FilterSyntaxError),
```
lbam, line 64:
```suggestion
        # XXX The baseId syntax shouldn't work with composite keys.
        # ("baseId", "REFERS REFERS/BASE BASE BASE/ID", FilterSyntaxError),
        # ("baseId", "REFERS REFERS/BASE BASE BASE/ID BASE/VOLGNR", FilterSyntaxError),
```

jjmurre (approved):
LGTM

PR #505, opened 2022-04-14
settings: replace os.getenv() with env.str() so proper error messages are given.

vdboor:
The settings `PUB_JWKS` and `OAUTH_JWKS_URL` are required in DSO-API.

If these are none, the error message will show a missing `JWKS` and `JWKS_URL` variable from authorization_django, which are not the correct setting names.


jjmurre (approved):
LGTM

vdboor:
breaks in CI.

PR #510, opened 2022-04-26
Remove FreeBSD installation instructions

lbam:
FreeBSD isn't the target platform, and the instructions consisted of a list of workarounds that no one is maintaining anymore.


jjmurre (approved):
LGTM. Let's clean this up avoiding the lava-flow antipattern :-)

PR #511, opened 2022-04-26
Rename FieldsToDisplay.__bool__ to FieldsToDisplay.reduced

lbam:
Explicit is better than implicit: this way, the method has a name indicating what it does, and we no longer need a reminder at the call site that we overloaded `__bool__` with a meaning that is not obvious. Also, PyCharm can find the method's call site, which it can't for `__bool__` without extensive type annotations.


vdboor (approved):
Prima! De naamgeving twijfelde ik nog even over, maar weet zelf ook geen betere.

PR #514, opened 2022-05-16
Filter refactor

vdboor:
This removes the dependency on django-filter, which is not well suited for our use-case.

The logic of django-filter requires us to predefine which filters exist, and how these are parsed. Such approach means all datasets have to construct huge filterset objects (all datasets * tables * fields * lookups) just to support query-string filters. Including relationship fields here only complicates matter further. Hence a new implementation.

* Remove old code
* Add new code
* Improved tests (taken from previous branch Lars worked on)


src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -0,0 +1,379 @@
...
+        if parent is None:
+            raise ValidationError(f"Field does not exist: {field_name}")
+
+        # Find the field by name, also handle workarounds for relation ID's.
+        field = _get_field_by_id(parent, name, is_last=(i == last_item))
+        if field is None:
+            raise ValidationError(f"Field does not exist: {field_name}")
+
+        if not user_scopes.has_field_access(field):
+            raise ValidationError(f"Access denied to filter on: {field_name}") from None
```
lbam, line 337: Shouldn't this be a `django.core.exceptions.PermissionDenied`?

src/dso_api/dynamic_api/filters/openapi.py
```diff
@@ -0,0 +1,177 @@
...
+    "gt": "Greater than; ",
+    "gte": "Greater than or equal to; ",
+    "not": "Exclude matches; ",
+}
+
+OPENAPI_LOOKUP_EXAMPLES = {
+    "isempty": "Whether the field is empty or not.",
+    "isnull": "Whether the field has a NULL value or not.",
+    "like": "Filter using wildcards (? and *).",
+    "in": "Filter using comma separated values: val1,val2,valN.",
```
lbam, line 56: These describe syntax, but not semantics.

```suggestion
    "like": "String match using wildcards (? and *).",
    "in": "Value must be one of a comma-separated list: val1,val2,valN.",
```

src/dso_api/dynamic_api/filters/backends.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Usage in views::
+
+        class View(GenericAPIView):
+            filter_backends = [filters.DynamicFilterBackend]
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """Filter the queryset of the view.
+        This hook function is called by Django REST Framework as apply
```
jjmurre, line 26: as apply this of -> to apply this?

src/dso_api/dynamic_api/filters/backends.py
```diff
@@ -0,0 +1,84 @@
...
+            filter_backends = [filters.DynamicOrderingFilter]
+    """
+
+    ordering_param = "_sort"  # Enforce DSO-specific name.
+
+    def get_ordering(self, request, queryset, view):
+        if self.ordering_param not in request.query_params:
+            # Allow DSO 1.0 Dutch "sorteer" parameter
+            # Can adjust 'self' as this instance is recreated each request.
+            if "sorteer" in request.query_params:
```
jjmurre, line 57: I thought the ideas was to use `_` prefixes for these parameters, or has this been changed?

src/dso_api/dynamic_api/filters/values.py
```diff
@@ -0,0 +1,136 @@
...
+    return Decimal(value)
+
+
+def str2isodate(value: str) -> date | datetime | None:
+    """Parse ISO date and datetime."""
+
+    # Try parsing full iso-8601 datetime.
+    result = parse_datetime(value)
+    if result is not None:
+        return result
```
lbam, line 42: `parse_datetime` can raise `ValueError` "if the input is well formatted but not a valid datetime". Is that accounted for?

src/dso_api/dynamic_api/filters/openapi.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    name = f"{prefix}{field.name}"
+    if field.relation and not field.is_loose_relation:
+        name += "Id"  # ForeignKey field
+    if lookup:
+        name = f"{name}[{lookup}]"
+    return name
+
+
+def _get_filter_description(field: DatasetFieldSchema, lookup=None) -> str | dict:  # noqa: C901
```
jjmurre, line 148: pipe notation for types is python3.10 only. Is mypy borking on this?

src/dso_api/dynamic_api/filters/values.py
```diff
@@ -0,0 +1,136 @@
+"""Conversion of query-string values to Python types."""
+from __future__ import annotations
+
+import math
+import re
+from datetime import date, datetime, time
+from decimal import Decimal
```
lbam, line 7: Why are we using `decimal` here?

src/dso_api/dynamic_api/filters/backends.py
lbam: This is for backwards compatibility with older versions of the DSO standard.

jjmurre:
I checked the PR out and ran the tests, all OK. However, on starting the DSO I get some errors and warnings. However, this could be data-related. 
```
ERRORS:
?: (models.E032) constraint name 'id_not_contains_slash' is not unique among models: afvalwijzer.afvalwijzer, bag.ligplaatsen, bag.nummeraanduidingen, bag.onderzoeken, bag.openbareruimtes, bag.panden, bag.standplaatsen, bag.verblijfsobjecten, bag.woonplaatsen, basiskaart.gebouwvlak, basiskaart.inrichtingselementlijn, basiskaart.inrichtingselementpunt, basiskaart.inrichtingselementvlak, basiskaart.labels, basiskaart.spoorlijn, basiskaart.terreindeelvlak, basiskaart.waterdeellijn, basiskaart.waterdeelvlak, basiskaart.wegdeellijn, basiskaart.wegdeelvlak, bbga.kerncijfers, bbga.statistieken, beheerkaartBasis.bgt, beheerkaartBasis.eigendomsrecht, beheerkaartBasis.kaart, beheerkaartCbsGrid.raster_10, beheerkaartCbsGrid.raster_100, beschermdestadsdorpsgezichten.beschermdestadsdorpsgezichten, borInspecties.monitorbeeldkwaliteit, bouwstroompunten.bouwstroompunten, brandkranen.brandkranen, brk.aantekeningenkadastraleobjecten, brk.kadastraleobjecten, brk.stukdelen, brk.tenaamstellingen, brk.zakelijkerechten, cmsa.sensor, duurzaamheid.energielabel, fietspaaltjes.fietspaaltjes, gebieden.bouwblokken, gebieden.buurten, gebieden.ggpgebieden, gebieden.ggwgebieden, gebieden.stadsdelen, gebieden.wijken, handelsregister.locaties, hoofdroutes.u_routes, huishoudelijkafval.adres_loopafstand, huishoudelijkafval.bag_object_loopafstand, huishoudelijkafval.bijplaatsingen, huishoudelijkafval.cluster, huishoudelijkafval.clusterfractie, huishoudelijkafval.container, huishoudelijkafval.containerlocatie, huishoudelijkafval.containertype, huishoudelijkafval.loopafstand_categorie, huishoudelijkafval.weging, indicatoren.buurt, indicatoren.ggw, indicatoren.stadsdeel, indicatoren.wijk, indicatoren.woonplaats, kwaliteitsmonitor.wonen, leidingeninfrastructuur.ligging_punt_totaal, meldingen.meldingen, parkeervakken.parkeervakken, standbedrijven.vestigingen, standvastgoed.gebouwen, stroomstoringen.stroomstoringen, varen.ligplaats, varen.opafstapplaats, wior.wior, woningbouwplannen.bag_pand_sloop_status, woningbouwplannen.gebied_bouwblok_woningen, woz.wozdeelobjecten, woz.wozobjecten.
?: (models.E032) constraint name 'identificatie_not_contains_slash' is not unique among models: brk.aantekeningenrechten, brk.gemeentes, brk.kadastralegemeentecodes, brk.kadastralegemeentes, brk.kadastralesecties, brk.kadastralesubjecten, haalcentraalbag.adresseerbareobjecten, haalcentraalbag.panden, haalcentraalbrk.kadasternatuurlijkpersonen, haalcentraalbrk.kadastraalonroerendezaken, meetbouten.meetbouten, meetbouten.metingen, meetbouten.referentiepunten, meetbouten.rollagen, nap.peilmerken.

WARNINGS:
datasets.Dataset: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the SchematoolsAppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
datasets.DatasetField: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the SchematoolsAppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
datasets.DatasetTable: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the SchematoolsAppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
datasets.Profile: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the SchematoolsAppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.

```

src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -0,0 +1,379 @@
...
+) -> DatasetFieldSchema | None:
+    """Find the field in the Amsterdam schema.
+    Handle workarounds for foreignkey ID fields that were previously exposed.
+    """
+    try:
+        return parent.get_field_by_id(name)
+    except SchemaObjectNotFound:
+        # Backwards compatibility: allow filtering on the "FOREIGNKEY_id" field,
+        # that was exposed in the API. This should only happen when the "Id" suffix is
+        # the truly a foreign key. (so avoid "streetId.number=..." or "anyfieldId=...").
```
lbam, line 370:
```suggestion
        # truly a foreign key. (so avoid "streetId.number=..." or "anyfieldId=...").
```

src/rest_framework_dso/filters/__init__.py
```diff
@@ -82,7 +82,6 @@ class View(GenericAPIView):
 
 .. _django-filter: https://django-filter.readthedocs.io/
 """
-from . import lookups  # noqa (import is needed for registration)
```
lbam, line 85: I think this whole file can go.

src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -0,0 +1,379 @@
...
+    "number": str2number,
+}
+
+FORMAT_PARSERS = {
+    "string": {
+        "date": str2isodate,
+        "date-time": str2isodate,
+        "time": str2time,
+        "uri": str,
+    }
```
lbam, line 52: Nitpick, but surely only `string` is ever going to have multiple formats, so we could hardcode that instead of looking it up in this dict.

src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -0,0 +1,379 @@
...
+from schematools.contrib.django.models import DynamicModel
+from schematools.exceptions import SchemaObjectNotFound
+from schematools.permissions import UserScopes
+from schematools.types import DatasetFieldSchema, DatasetTableSchema
+from schematools.utils import to_snake_case
+
+from .values import str2bool, str2geo, str2isodate, str2number, str2time
+
+# Allow notations: field.subfield[lookup]
+RE_KEY = re.compile(r"\A(?P<path>[\w\d_-]+(?:\.[\w\d_-]+)*)(?:\[(?P<lookup>[a-z0-9A-Z_-]+)])?\Z")
```
lbam, line 28: There's a `parse_filter` function in `dso_api.dynamic_api.permissions` to parse a filter into field name and lookup operator. It produces specific error messages and is (IMHO) more readable than this.

src/dso_api/dynamic_api/filters/values.py
```diff
@@ -0,0 +1,136 @@
...
+        return False
+    else:
+        raise ValueError("expect true/false")
+
+
+def str2number(value: str) -> Decimal:
+    if not RE_DECIMAL.match(value):
+        raise ValueError("expecting number")
+
+    return Decimal(value)
```
lbam, line 33: Just let Python do the parsing and check for infinities afterward, it produces better error messages and is probably faster:

```suggestion
    if "." in value:
        value = float(value)
        if not math.isfinite(value):
            raise ValueError(f"not a valid number: {value!r}")
        return value
    return int(value)
```

src/dso_api/dynamic_api/filters/openapi.py
vdboor: nope! 🤩 because the file has `from __future__ import annotations` in it! :-D

src/dso_api/dynamic_api/filters/values.py
vdboor: JavaScript number has a higher precision then a Python float.

src/dso_api/dynamic_api/filters/values.py
vdboor: I'm not sure.. I wanted to make sure the input can only be a simple decimal notation, and not `0.321000e+2` for example. That would still work in such case.

src/dso_api/dynamic_api/filters/parser.py
vdboor: I found out about that function later too. I'm not sure about the `permissions` module. We'll likely have to refactor that indeed. All filters and views already check for permissions now, so it looks like an additional "firewall" upfront to perform the same checks.

src/rest_framework_dso/filters/__init__.py
vdboor: indeed, thanks! this error happened as I was splitting up commits.

src/dso_api/dynamic_api/filters/values.py
```diff
@@ -0,0 +1,136 @@
...
+from django.utils.translation import gettext_lazy as _
+from gisserver.geometries import CRS
+from rest_framework.exceptions import ValidationError
+
+# Don't want Decimal("NaN"), Decimal("-inf") or '0.321000e+2' to be accepted.
+RE_DECIMAL = re.compile(r"^[0-9]+(\.[0-9]+)?$")
+
+
+def str2bool(value: str) -> bool:
+    value = str(value).lower()
```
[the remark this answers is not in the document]
vdboor: This was there to handle unit tests that passed querystring-dicts with boolean values in. Should I remove this, or leave it? because this does fix such corner case.

src/dso_api/dynamic_api/filters/openapi.py
vdboor: ok, good point. I've updated this to:

```
    "like": "Matches text using wildcards (? and *).",
    "in": "Matches any value from a comma-separated list: val1,val2,valN.",
```

src/dso_api/dynamic_api/filters/parser.py
vdboor: tx!

src/dso_api/dynamic_api/filters/parser.py
vdboor: In OpenAPI, integer and number can also have a `format` argument. So this is more to be generic and exact.

src/dso_api/dynamic_api/filters/values.py
lbam: According to the [ECMAScript spec](https://262.ecma-international.org/11.0/), JavaScript numbers are IEEE 754 double precision (64 bits). Python doesn't guarantee anything, but CPython uses C's `double` type, which is also IEEE 754 double precision, on all major platforms.

src/dso_api/dynamic_api/filters/values.py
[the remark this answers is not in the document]
lbam: Sounds like the unit test should be fixed instead.

src/dso_api/dynamic_api/filters/parser.py
lbam: That sounds better. In JSON Schema, I don't think `number` can have a format.

src/dso_api/dynamic_api/filters/values.py
lbam: Yes, but my point is that `parse_datetime` is not inside a `ValueError` block but the code below is...

src/dso_api/dynamic_api/filters/openapi.py
lbam: Even better!

lbam (dismissed):

src/dso_api/dynamic_api/filters/values.py
vdboor: ok, fixed!

src/dso_api/dynamic_api/filters/values.py
vdboor: chosen to leave as it is

src/dso_api/dynamic_api/filters/parser.py
vdboor: chosen to leave as it is for now

src/dso_api/dynamic_api/filters/values.py
vdboor: chosen to leave as it is.

PR #518, opened 2022-05-23
Fixes for filtering with field[in]=.. and M2M paths

vdboor:
Fix on top of the new filter refactor, with added tests.

* Fix filtering on `field[in]=value1,value2`.
* Allow `[in]` lookup on nearly all field types, except geometry
* Fix M2M results, add `SELECT DISTINCT`.


lbam (approved):
I only had a quick look at this, but the unit test changes look convincing.

PR #533, opened 2022-06-28
Get better diagnostics for missing API keys in remotes

lbam:
We want "Missing API Key" from the remote, not "Invalid API Key",
when we can't send the key.


jjmurre (approved):
LGTM

lbam:
Hold on, some tests are failing because they need the key to be set.

jjmurre (approved):
✅

PR #549, opened 2022-07-14
Split dynamic_api/serializers.py into factory and runtime serializer part

vdboor:
Literal code split, no changes except for documentation improvements.


lbam:
What is the benefit of this?

PR #552, opened 2022-07-20
Hide fields in _links when there is no access for it (e.g. display field)

vdboor:
This reorganizes the serializers, so permission checks can be applied to the serializer `_links` elements,

Commit 1 reorganizes the `LinksSerializer` class hierarchy so it no longer has to disable functionality from the base class.
Commit 2 inserts the checks and deals with the side effects from that.

Lastly, when a link item would not return any fields, it will omit the entry completely.

### DIFF
--- a/src/dso_api/dynamic_api/serializers/base.py
+++ b/src/dso_api/dynamic_api/serializers/base.py
@@ -204,6 +204,8 @@ class DynamicListSerializer(DSOModelListSerializer):
 class FieldAccessMixin(serializers.ModelSerializer):
     """Mixin for serializers to remove fields the user doesn't have access to."""
 
+    fields_always_included = set()
+
     @cached_property
     def _request(self):
         """Get request from this or parent instance."""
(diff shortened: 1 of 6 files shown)


src/dso_api/dynamic_api/serializers/base.py
```diff
@@ -259,6 +257,12 @@ def _apply_permission(self, field_name: str, field: serializers.Field) -> bool:
             # Check access
             permission = user_scopes.has_field_access(field_schema)
             if not permission:
+                logging.debug(
```
lbam, line 260:
```suggestion
                logging.info(
```

This is security-related. In fact, I suspect we should audit-log it, but we can leave that out of this PR.

src/dso_api/dynamic_api/serializers/base.py
```diff
@@ -234,21 +214,39 @@ def _request(self):
...
+            field.Meta.model.table_schema()
+        ):
+            # Extra case for the _links section: when a LinkSerializer object will not render
+            # any fields (due to not having access to the table), don't render that block at all.
+            # Otherwise, it would just render `"relation": {}`. Now the field is simply missing,
+            # in the same way other body fields are omitted when there is no access.
+            #
+            # When a user has access to a relation that is NULL, it will still be rendered
+            # as "relation": null, or "relation": [] for M2M, so there is no ambiguity here.
+            logging.debug(
```
lbam, line 244:
```suggestion
            logging.info(
```

Again, security-related.

src/tests/test_dynamic_api/test_serializers.py
```diff
@@ -625,7 +626,54 @@ def test_indirect_self_reference(ligplaatsen_model, filled_router):
...
+        ContainerSerializer = serializer_factory(afval_container_model)
+        container_serializer = ContainerSerializer(
+            afval_container, context={"request": drf_request}
+        )
+        data = normalize_data(container_serializer.data)
+        assert data == {
+            "_links": {
+                # The cluster block is completely omitted because there is no access to the table.
+                # So won't expose the identifier either. This isn't ambiguous either,
+                # because if the relation was missing, it would be "cluster": null,
```
lbam, line 640: Reconstruction of a comment that I just posted, but that GitHub seems to have swallowed:

```suggestion
                # Just showing the href would leak the identifier. Setting "cluster": null
                # would be ambiguous, since that already means "no cluster for this container".
```

lbam (changes requested):

src/tests/test_dynamic_api/test_serializers.py
vdboor: thanks! I was struggling with that last commit all the time.

lbam (approved):

PR #566, opened 2022-07-27
Clean up URL construction in dynamic_api.routers

lbam:
One to_snake_case call down, many more to go.


Yrob (approved):
another one bites the dust

PR #568, opened 2022-07-28
AB#50108: Support filtering/joining of loose relations

vdboor:
This change allows loose relations to be filtered like any other field. 

Since `LooseRelationField` became a `ForeignKey`, some fixes were applied to DSO-API. Next, the filtering was improved to handle joins. This needed access to the `TemporalTableQuery` inside `QueryFilterEngine`. Since the first needed a `Request` object, that construction has to be changed too. As bonus, `related_field_ids` was improved in schematools, which removes the need for various hacks.

This builds on top of Amsterdam/schema-tools/pull/308


src/dso_api/dynamic_api/temporal.py
```diff
@@ -166,13 +224,23 @@ def url_parameters(self) -> dict[str, str]:
             return {}
 
 
+def get_request_date(request: Request) -> datetime:
+    """Find the temporal query date, associated with the request object."""
+    try:
+        # Make sure all queries use the exact same now, not seconds/milliseconds apart.
```
lbam, line 230: This is mainly for SQL, right? Can't we use the Postgres [`NOW`](https://www.postgresql.org/docs/14/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT) function instead? That returns the starting time of the transaction.

src/dso_api/dynamic_api/temporal.py
```diff
@@ -166,13 +224,23 @@ def url_parameters(self) -> dict[str, str]:
...
+        return request._now
+    except AttributeError:
+        request._now = now()
```
lbam, line 233: If we do want to keep this, I think it might be cleaner to set it from some piece of middleware. Maybe the AuthMiddleware could be expanded to become an `AdornMiddleware` that sets various bits of context info on the request.

src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -173,11 +175,24 @@ class QueryFilterEngine:
         "sorteer",
     }
 
-    def __init__(self, user_scopes: UserScopes, query: MultiValueDict, input_crs: CRS):
+    @classmethod
+    def from_request(cls, request):
```
lbam, line 179: Named constructors FTW!

src/dso_api/dynamic_api/temporal.py
lbam: Not crucial though, feel free to ignore.

lbam (dismissed):

src/dso_api/dynamic_api/temporal.py
vdboor: Nope. All the queries for the request should use the same date. Otherwise, there is a 10-100ms difference in the requested date.

lbam: Do the queries not run within a transaction?

vdboor: Interesting about `NOW`, didn't know that. But that will be tricky, and I don't think there is a transaction at all.

PR #585, opened 2022-09-12
Allow switching the database role based on end-user context

vdboor:
For internal employees, it will fallback to a separate internal role. This functionality is only enabled on Azure.


src/dso_api/dbroles.py
```diff
@@ -0,0 +1,143 @@
...
+            if is_internal(user_email):
+                # BBN1: Internal employee, no specific account
+                cls._set_role(user_connection, INTERNAL_ROLE, user_email)
+            else:
+                logger.exception("External user %s has no database role %s", user_email, role_name)
+                raise PermissionError(f"User {user_email} is not available in database") from e
+
+    @classmethod
+    def _set_role(cls, user_connection: BaseDatabaseWrapper, role_name: str, app_name: str):
+        # By starting a transaction, any collection pooling (e.g. PgBouncer)
```
DonQueso89, line 113:
```suggestion
        # By starting a transaction, any connection pooling (e.g. PgBouncer)
```

src/dso_api/dbroles.py
```diff
@@ -0,0 +1,143 @@
...
+                raise
+            else:
+                setattr(user_connection, "_active_user_role", role_name or "NONE")
+                logger.info("Activated end-user database role '%s' for '%s'", role_name, app_name)
+
+    @classmethod
+    def deactivate_end_user(cls):
+        """Rollback the transaction when the local user was activated."""
+        user_email = getattr(cls.current_user, "email", None)
+        if not user_email or not setattr(default_connection, "_active_user_role", None):
```
DonQueso89, line 133: shouldnt this be `getattr`? otherwise it always evaluates to True

DonQueso89 (dismissed):

src/dso_api/dbroles.py
vdboor: yikes. indeed. I've wrapped all getattr() logic in separate functions now,

DonQueso89 (approved):

PR #592, opened 2022-09-19
Give 400 instead of 500 for "Related Field got invalid lookup"

lbam:
Requests such as

    /v1/meldingen/statistieken/?buurt.identificatie[like]=X

would give error 500 due to Django not being able to handle them. Give a 400 instead.


vdboor (approved):

vdboor:
Clearly the `[like]` needs to happen against the relation_id field. But this is useful to add regardless of that.

PR #598, opened 2022-10-19
Update DSO-API with new schematools release changes

vdboor:
This fixes the shortnames appearing in the API.

The cause of this bug happened deep inside schematools, where the `field.name` would return the database `shortname`. Thus, model fields, and even API serializer fields received that shortname. This PR brings DSO-API in line with the changes in schematools.

### DIFF
--- a/src/dso_api/dynamic_api/filters/parser.py
+++ b/src/dso_api/dynamic_api/filters/parser.py
@@ -9,7 +9,8 @@ from datetime import datetime
 from functools import reduce
 from typing import Any, NamedTuple
 
-from django.core.exceptions import ValidationError as DjangoValidationError, FieldError
+from django.core.exceptions import FieldError
+from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db import models
 from django.db.models import Q
 from django.utils.datastructures import MultiValueDict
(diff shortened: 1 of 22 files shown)


src/dso_api/dynamic_api/filters/parser.py
```diff
@@ -440,25 +441,27 @@ def _to_orm_path(parts: list[FilterPathPart]) -> str:
...
 
+    @property
+    def name(self):
+        # Simulates the same properties as the DatasetFieldSchema provides.
+        return toCamelCase(self.id)
+
+    @property
+    def python_name(self):
+        # Likewise, an exact replica of what DatasetFieldSchema provides.
+        return to_snake_case(self.id)
```
lbam, line 463: If we put the casing logic on `AdditionalRelationSchema` as well, we can do:

```suggestion
    @property
    def name(self):
        return getattr(self.reverse_field or self.field, "name")

    @property
    def python_name(self):
        return getattr(self.reverse_field or self.field, "python_name")
```

That's -1 import of these functions.

vdboor: thanks, great fix! I was already puzzling with that one...

lbam (approved):

lbam (approved):

PR #602, opened 2022-10-24
Bump requirements for security, future maintenance and space for Django 4

vdboor:
Bump all requirements for  security, future maintenance and to make room for future Django upgrades.

Django is kept at 3.2 for LTS support, but running this code on Django 4.0 works.

For Django 4.1 tests to pass, the ChunkedQuerysetIterator needs to be removed because Django 4.1 does the same internally.


lbam:
Good idea, but we get a smaller patch if we merge #600 first.

lbam (approved):

PR #603, opened 2022-10-24
Test isnull filter on a relation field

lbam:
Inspired by a case reported on Slack where this doesn't work. Won't fix the issue, but makes it easier to spot regressions.


jjmurre (approved):
LGTM

PR #611, opened 2022-11-08
Generate docs on the fly

lbam:
This adds an endpoint /v1/docs/datasets/{dataset_id}.html that displays automatically generated HTML documentation for a dataset. The contents match those of the Sphinx documentation, which these docs must ultimately replace. There is also an overview of datasets at /v1/docs/.

The HTML isn't styled yet, but the functionality is there. For now, this will not display on api.data.amsterdam.nl/v1/docs/, which uses proxy rules to display docs from a separate Docker image instead.

For [AB#56614](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/56614).

### DIFF
--- a/src/dso_api/dynamic_api/routers.py
+++ b/src/dso_api/dynamic_api/routers.py
@@ -24,7 +24,7 @@ from __future__ import annotations
 
 import logging
 from itertools import chain
-from typing import TYPE_CHECKING, Iterable
+from typing import TYPE_CHECKING, Iterable, Iterator
 
 from django.apps import apps
 from django.conf import settings
@@ -45,6 +45,7 @@ from .serializers import clear_serializer_factory_cache
 from .utils import get_view_name
 from .views import (
     APIIndexView,
+    DatasetDocView,
     DatasetMVTSingleView,
     DatasetMVTView,
     DatasetWFSView,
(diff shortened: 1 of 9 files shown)


vdboor (approved):

src/dso_api/dynamic_api/routers.py
```diff
@@ -332,6 +338,15 @@ def _build_remote_viewsets(self, api_datasets: Iterable[Dataset]):
 
         return tmp_router.registry
 
+    def _build_doc_views(self, datasets: Iterable[Dataset]) -> Iterator[URLPattern]:
+        for dataset in datasets:
```
jjmurre, line 342: Moeten we nog rekening houden met `status == 'niet_beschikbaar'` bij generatie van de docs?

jjmurre: En we hebben ook nog de `enable_api` vlag.

lbam: Gebeurt al waar deze methode wordt aangeroepen:

        # Create viewsets only for datasets that have an API enabled
        db_datasets = [ds for ds in db_datasets if ds.enable_api]
        dataset_routes = self._build_db_viewsets(db_datasets)
        # Same for remote API's
        api_datasets = list(get_active_datasets().endpoint_enabled())
        remote_routes = self._build_remote_viewsets(api_datasets)

        datasets = db_datasets + api_datasets

        doc_urls = self._build_doc_views(datasets)

jjmurre (approved):

PR #623, opened 2022-11-15
Remove test for old Azure Blob Storage support

lbam:
The corresponding feature was removed in #600, but I missed this test since it kept succeeding.


jjmurre (approved):
LGTM

PR #629, opened 2022-11-22
Port generic docs to HTML generator via Markdown

lbam:
Existings docs translated using `pandoc -t gfm`.

The screenshots were copied from their location in docs. They're unmodified, so Git deduplicates them.


jjmurre (approved):
LGTM

PR #639, opened 2022-11-30
Prettify generated docs with CSS

lbam:
Copy-pasted from the dataset query UI, so everything looks similar.

Still figuring out how to style the Markdown-generated docs.


jjmurre (approved):
LGTM

PR #660, opened 2023-01-11
Fix invalid CRS handling

lbam:
Uses the new version of gisserver, which no longer raises `SyntaxError`. Only check for its `ExternalValueError` type, so we don't send error 406 for `ValueErrors` caused by bugs.

Fixes [AB#67466](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/67466).


jjmurre (approved):
LGTM

PR #668, opened 2023-01-17
Docs: log and ignore broken datasets

lbam:
`/v1/docs` is broken because of one dataset. I can't reproduce the error locally. For now, log the error so we can serve the rest of the list.


DonQueso89 (approved):
if this is not fixed soon, it needs a fat comment

lbam:
I intend to work on this tomorrow. The broken dataset needs fixing. Thanks for the review!

PR #671, opened 2023-01-24
Set uwsgi buffer size to max

lbam:
Should fix issue where headers are too large for buffer.


jjmurre (approved):
LGTM, let's see what happens.

PR #677, opened 2023-01-25
Minor refactoring of doc generating code

lbam:
Get filters from the filter parser instead of listing them again. The only inconsistency I found was boolean missing the isnull lookup, but there might be more.

Merged `is_relation` and `is_foreign_id` in input to template processing.


jjmurre (approved):
LGTM

PR #685, opened 2023-01-31
Fix serving of static files

lbam:


jjmurre (approved):
LGTM, Nice you found it!

PR #693, opened 2023-02-02
Restore search in docs

lbam:
Could use some (a lot of) tweaking, but works. Includes a vendored fuse.js 6.6.2 (https://fusejs.io/).


jjmurre (dismissed):
Nice!

jjmurre (approved):
👍🏼

PR #699, opened 2023-02-08
Rewrite Haal Centraal BAG proxy

lbam:
The new proxy is a simple Django view, decoupled from DRF's serializers. It passes through responses almost unchanged, so it cannot convert from JSON to other formats like CSV.

The schema for haalcentraal/bag should be removed for this to work: Amsterdam/amsterdam-schema#624. For [AB#69558](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/69558).


src/dso_api/dynamic_api/remote/views.py
```diff
@@ -23,6 +28,55 @@
...
+
+    This is simple pass-through proxy: we change the url, send the request on,
+    then fix up the response so that self/next/prev links point to us instead of HC.
+    """
+
+    _BASE_URL = settings.DATAPUNT_API_URL + "v1/haalcentraal/bag/"  # XXX use reverse?
+
+    def __init__(self):
+        super().__init__()
+        self.endpoint = "https://api.bag.acceptatie.kadaster.nl/esd/huidigebevragingen/v1/"
```
jjmurre, line 64: Why not use the variable from settings for self.endpoint?

lbam: Because I forgot to update this. Fixed now.

jjmurre (approved):
LGTM

PR #701, opened 2023-02-09
Simplify regexp separator splitting

lbam:


src/dso_api/dynamic_api/utils.py
```diff
@@ -14,13 +14,12 @@
 
 from rest_framework_dso.fields import GeoJSONIdentifierField
 
-PK_SPLIT = re.compile("[_.]")
+AFTER_SEP = re.compile("[^_.]*$")
 
 
-def split_on_separator(value):
+def split_on_separator(value: str):
```
jjmurre, line 20: This is not going to work, because the return value should be a list of 2 elements.  So, for `foo_bar_baz` the return value should be `['foo_bar', 'baz']`. This is older code, it would be nice if the signature of the return value `-> list[str]` could also be added.

lbam: You're right, I forgot to check in some bits. Fixed now.

lbam:
Closing. Poorly thought through.

PR #705, opened 2023-02-10
Restore HTTP error handling for haalcentraal/bag + further refactor of Haal Centraal client

lbam:
See commit messages for details.

### DIFF
--- a/src/dso_api/dynamic_api/remote/clients.py
+++ b/src/dso_api/dynamic_api/remote/clients.py
@@ -42,6 +42,109 @@ class RemoteResponse:
     data: Union[list, dict]
 
 
+def call(pool: urllib3.PoolManager, url: str, **kwargs) -> HTTPResponse:
+    """Make an HTTP GET call. kwargs are passed to pool.request."""
+
+    host = urlparse(url).netloc
+    t0 = time.perf_counter_ns()
+    try:
+        # Using urllib directly instead of requests for performance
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/remote/clients.py
```diff
@@ -42,6 +42,109 @@ class RemoteResponse:
...
+        response.status,
+        response.reason,
+        (time.perf_counter_ns() - t0) * 1e-9,
+    )
+
+    if response.status >= 200 and response.status < 300:
+        return response
+
+    # We got an error.
+    if logger.isEnabledFor(logging.DEBUG):
```
jjmurre, line 83: Is the `if` necessary? Because the logger handles this for you depending on the log level.

lbam: This is ancient code from 2020, see ab346a13764c31e6b804f3b8d11708708b079f31. It looks overly cautious to me too, I'll see if I can get rid of it.

lbam: I think the issue is that `.data` is a `property` that parses and caches the response, which we don't want to do except if we actually want to print the message.

jjmurre: Ah, makes sense.

jjmurre (approved):
LGTM

PR #708, opened 2023-02-13
Generalize _rewrite_links in remote.views

lbam:
For Haal Centraal BRK, we need something more general, because its output can contain both absolute and relative URLs, and we need to rewrite both.


lbam:
Also, this is cleaner in that we can actually parse the URL in the callback if we ever need to do that. The old implementation assumed we could always do a simple string replace.

Yrob (approved):

PR #712, opened 2023-02-14
Fix documentation link in OpenAPI spec

Yrob:
Documentation link was pointing to production independent of environment. Now we use the DATAPUNT_API_URL envronment setting, so the OPENAPI spec in acceptance points to acceptance docs

> Don't forget about...
> * Tests
> * Documentation in `dev-docs/`
>


jjmurre (approved):
LGTM

PR #719, opened 2023-02-20
Simplify filter implementation

lbam:
An empty `django.db.models.Q` object is a no-op and is False in a boolean context. Use that to our advantage: we don't need to build up lists of `Q`'s with special handling for empty lists. Only `QueryFilterEngine.filter_queryset` needs to check whether it needs to do any work.

I also removed `LOOKUP_MERGE_OPERATORS`, which was just a complicated way of checking for `not`.


jjmurre (approved):
👍🏼

Yrob (approved):

PR #727, opened 2023-03-08
Fix publishers display in docs

lbam:
Locally, publishers would display as

    {'name': 'Datateam Stedelijke Ontwikkeling en Beheer', 'id': 'SOEB', 'shortname': 'soeb', 'tags': {'team': 'stedelijke ontwikkeling en beheer', 'costcenter': '738000002.4810.13519'}, 'type': 'publisher'}

while on PROD, they would display as

    {'$ref': 'publishers/SOEB'}

Try to expand the publishers. Needs a workaround for the unit tests, where the loader on a dataset schema is apparently reset to an instance of the base class SchemaLoader. I've no idea what's going on there.


jjmurre (approved):
LGTM.

PR #734, opened 2023-04-03
Tiny optimizations for HAL-JSON rendering

lbam:
When rendering a dict, we've already checked if it's empty, so we can save a single yield by putting the `'{'` in the first separator.

Also, the `level` variable can be a boolean, because we only increment it and check for zero. I didn't benchmark this, but we've had a case before (Amsterdam/django-gisserver#11) where just keeping track of a count inside serialization code caused a 5% throughput drop.


barrydaniels-nl (approved):
Looks good

PR #736, opened 2023-04-05
MVT: less detail at higher zoom levels

lbam:
This should alleviate some of the memory problems of our MVT endpoints, by simply not sending non-geo fields' data at high zoom levels. The cutoff point is hardcoded, based on some experimentation with MVT maps. At level 15, street names should be legible and the city of Amsterdam consists of many separate tiles, so the load of serving it can be spread among DSO-API instances.

For [AB#61546](https://dev.azure.com/CloudCompetenceCenter/089b38ee-0066-4b66-a36c-20744a0e4348/_workitems/edit/61546).

### DIFF
--- a/src/dso_api/dynamic_api/views/mvt.py
+++ b/src/dso_api/dynamic_api/views/mvt.py
@@ -114,6 +114,7 @@ class DatasetMVTView(CheckPermissionsMixin, MVTView):
     def get(self, request, *args, **kwargs):
         kwargs.pop("dataset_name")
         kwargs.pop("table_name")
+        self.z = kwargs["z"]
 
         t0 = time.perf_counter_ns()
         result = super().get(request, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


src/dso_api/dynamic_api/views/mvt.py
```diff
@@ -114,6 +114,7 @@ def setup(self, request, *args, **kwargs):
     def get(self, request, *args, **kwargs):
         kwargs.pop("dataset_name")
         kwargs.pop("table_name")
+        self.z = kwargs["z"]
```
jjmurre, line 117: No risk for KeyError here?

jjmurre (approved):
LGTM. One question.

src/dso_api/dynamic_api/views/mvt.py
lbam: The URL format is fixed in this line:

https://github.com/Amsterdam/dso-api/blob/aaa702c82d3acb575d4629a0f4a00c1f10455a23/src/dso_api/dynamic_api/routers.py#L390

PR #738, opened 2023-04-10
More micro-optimizing on JSON rendering

lbam:
Bytes concatenation `x+y` is about 10x faster than `b"%b%b" % (x,y)`.

`sep + orjson.dumps({key: value})` is a bit faster than formatting. We now need to construct a `dict`, but we no longer need to construct a `tuple`, so the memory allocation costs are similar.

`HALJSONRenderer.render_json` is now some 6% faster on my (ad hoc) microbenchmark.


jjmurre (dismissed):
LGTM.

jjmurre (approved):

PR #742, opened 2023-05-02
Bump orjson

lbam:
3.8.10 release notes mention a compatibility problem with CPython 3.12 (alpha version) in older releases.


barrydaniels-nl (approved):
Top

PR #743, opened 2023-05-09
Snake-case dataset ids for linking to docs

lbam:
E.g., https://api.data.amsterdam.nl/v1/bor_inspecties/ would previously show a link to /v1/docs/datasets/borInspecties.html, which does not exist. The dataset docs use the snake-case form, just like the endpoints.

Also fixed a typo.


barrydaniels-nl (approved):
Looks good

PR #751, opened 2023-07-12
Add API Key Middleware

jjmurre:
For now, the API Key has been set to be optional.
However, if provided, it will be validated.

pip-tools needed to be upgraded to get it running again. Django check gave a warning about the Middleware not being in the correct order in the settings.MIDDLEWARE, so it had been moved.


barrydaniels-nl (approved):
Ok

PR #766, opened 2023-09-25
Bump schematools to 5.17.3

jjmurre:
To get the fix in the `--create-schemas` option of the `create-tables` mgm command.


roelof-adam (approved):
lgtm

PR #769, opened 2023-10-04
Temporary removal of link to api key server.

jjmurre:
Because the azure production environment is not yet ready, we commented out the link to the apikey server form.


barrydaniels-nl (approved):
Ok

PR #775, opened 2023-10-05
Make useage of api keys configurable.

jjmurre:
It can be annoying that the api key middelware keeps trying to get the signing keys when developing the dso locally.

So, now an environment variable `APIKEY_ENABLED` can be used to avoid the activation of the api key middleware. If the value of `APIKEY_ENABLED` is `false`, the middleware is not activated.


barrydaniels-nl (approved):
Ok

PR #780, opened 2023-10-25
Re-add the link to the api key form.

jjmurre:
Now that the api key server is running in PRD
we can re-add the link to the form where a key can be requested.


barrydaniels-nl (approved):
Ok

PR #787, opened 2023-12-06
AB#94572  Add download button to API page

jjmurre:
Normally, the output (be it json, csv or geojson) is rendered in the html page. Using `_format=<a format>` the use can start downloading to a file. However, when an `Authorization` header is needed, this cannot be added to this download request (only when using curl, which is a bridge too far for some users).

A download button had been added to download data *and* take headers and query parameters into account.

> Don't forget about...
> * Tests
> * Documentation in `dev-docs/`
>


lennardstam (approved):
OK

PR #790, opened 2024-01-08
AB#101824 fix partially protected relations

jjmurre:
The Django ORM has been designed around the idea that a single database user with wide permissions is accessing the database. Authorization-based filtering is done at the python level, e.g. in DRF serializers.

However, our PostgreSQL database has been protected at the row-level with fine-grained permissions and the user accessing the database is not a generic database user, but the user accessing the API (based on the JWT auth token).

This leads to problems, especially when following relations. The Django ORM tries to fetch all rows (SELECT * FROM) and a permission exception will be raised by the database driver.

At several places extra scope-checks need to be applied to force the ORM to only fetch the rows that are allowed.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -10,8 +10,8 @@ sync:
 
 .PHONY: requirements
 requirements: requirements.in requirements_dev.in
-	pip-compile -v --generate-hashes --resolver=backtracking --output-file requirements.txt requirements.in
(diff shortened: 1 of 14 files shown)


gurustacks (approved):
Ok

barrydaniels-nl (approved):
Ok

PR #795, opened 2024-01-17
AB#104026 Fix viewset name for docs

jjmurre:
To generate the viewnames for the documentation, the db_name_variant function was used. However, that takes shortnames into account. The Django viewset ids names are using the full names of the schema-tables, so there is a mismatch leading to a 500 error for datasettables with a shortname.


barrydaniels-nl (approved):
Ok

PR #810, opened 2024-02-23
AB#107499 Fix issue with expand query with only open fields

jjmurre:
It should be possible for a request to fetch unprotected fields (using the `_fields` query parameter).


barrydaniels-nl (approved):
Looks good

PR #830, opened 2024-04-29
Remove setting that causes disabling of geosearch for datasets.

jjmurre:
This was an old setting, because bag was being geosearched using the old bag/v11 until recently. But now, we want geosearch to use the ref db for Bag.


roelof-adam (approved):
OK

PR #834, opened 2024-06-06
Make bulk export endpoint variable

Yrob:
Fixed bulk export endpoint. And made it configurable with variable


RobbieLuchs (approved):
lgtm

PR #839, opened 2024-06-17
Fix API performance

vdboor:
Fix massive DSO-API performance issue, due to deepcopy()

DRF performs a deepcopy() of the serializer._declared_fields before it will handle the request. This is analogous to how Django forms work, and updating the fields with request state and 'parent' references.

Since the deepcopy() also touched dataset definitions of Amsterdam Schema, it basically took 14sec to copy all those dicts as well. This can be avoided, bringing to performance of DSO-API back to reasonable levels.


barrydaniels-nl (approved):
OK

PR #845, opened 2024-07-10
Let dependabot check for package updates

Yrob:
Create dependabot yml file configuring dependabot for weekly updates. It should run pip compile, update all packages and create a PR.


vdboor (approved):
Nice!

PR #846, opened 2024-07-11
Upgrade to amsterdam-schema-tools 6.0 for better performance

vdboor:
This improves the performance of has_field_access() checks and direct "auth" field access.

As part of this improvement, the call signature of has_all_scopes() was changed, which affected the old Haalcentraal BRK API proxy code.


barrydaniels-nl (approved):
Ok

PR #848, opened 2024-07-18
Django 3.2 -> 4.2 upgrade

vdboor:
This bumps all packages, and fixes minor test regressions for Django 4 / MVT tiling

* Moved the iterator code together as it needed
* The iterator optimization needed a different way to hook into Django.
* MVT tiling responses no longer use `"type": 1` but `"type": "FeatureCollection"`.


barrydaniels-nl (approved):
LGTM

PR #849, opened 2024-07-18
Fix timezone passing to datetime, since Django 4.2 uses zoneinfo

vdboor:
This ruff fix previously didn't work on Django 3.2 as that uses pytz. Passing such timezone to datetime causes +00:18 min shifts instead of the expected +01:00 change.

This is caused by the way pytz handles the zoneinfo from history and needs pytz.timezone.localize() to create the proper timezone information. See: https://groups.google.com/g/django-users/c/rXalwEztfr0/m/QAd5bIJubwAJ


barrydaniels-nl (approved):
LGTM

barrydaniels-nl (approved):
Ok

barrydaniels-nl (approved):
Ok

PR #865, opened 2024-08-01
Some improvements to the dev experience

Yrob:
* Fix the startup order in the compose file, so the application does not start before database and redis are ready.
* Add mock data option to compose file. Use the MOCK_DATA´ env variable to fill the local db with mock data.
* Unset allowed opener header so auth callback works again
* Update install instructions in Readme


barrydaniels-nl (approved):
Nice!

PR #866, opened 2024-08-01
Cleanup filter tests, move to single file

vdboor:
It turned out there were multiple places were filtering tests happened.
The mandatory filter tests I wrote back then had a lot of code in a single test. This is now split into separate tests, and parametrized tests to make this easier to read and extend.


barrydaniels-nl (approved):
OK

PR #895, opened 2024-10-14
Make sure error messages don't get JSON syntax-highlight formatting

vdboor:
This simply writes the HTML content of the 500 server error as received source text.


barrydaniels-nl (approved):
ok

PR #899, opened 2024-10-24
Fix database error message handling

vdboor:
* Report when database roles have no access (no longer gives a 500)
* Fix 500 error HTML response messing up JSON syntax highlight.
* Make error handling easier, no longer jump through hoops to have a 404 for invalid page numbers.
* Remove server-side cursors as that breaks with PgBouncer.


barrydaniels-nl (approved):
Ok

PR #901, opened 2024-10-29
Added OpenTelemetry, remove OpenCensus.

vdboor:
Also changed the log formatting to use an actual JSON-formatter. The old solution generated JSON-like log strings by injecting the strings into an existing JSON-formatted string. This doesn't escape the '"' character, thus breaks with some django server messages.

Fixed requirements file (regenerating it broke, sorting the requirements out fixed it).


barrydaniels-nl (approved):
ok

PR #924, opened 2025-01-08
Upgrade python to 3.13

jasper-jager:
After running `make upgrade` the issues with rpds_py were fixed. Upgrading to python 3.13 required running pip-compile with `--allow-unsafe`. See: https://github.com/jazzband/pip-tools?tab=readme-ov-file#deprecations


vdboor (approved):
LGTM!

Lijkt me goed om deze als 1 squash commit te mergen?

PR #930, opened 2025-01-20
Update schema tools to fix mock data

jasper-jager:
When running docker compose with mock data the command failed due to an outdated schema being provided. We now use the active Amsterdam schema to generate mock data.

The provided schema container is now also obsolute and thus removed.


barrydaniels-nl (approved):
ok

PR #962, opened 2025-03-03
Fix xss vulnerability in browsable api

Yrob:
User input was not properly sanitized before writing to DOM.
This should fix an issue brought up by github code-scanning.


barrydaniels-nl (approved):
lgtm

PR #982, opened 2025-03-31
Remove the remote proxy support from DSO-API

vdboor:
This functionality has been replaced by the Amsterdam/haal-centraal-proxy repository, which does a better tailor made implementation with extra business logic. The existing DSO-API logic is no longer used, and only complicates the codebase at this point.


jasper-jager (dismissed):
Ok!

jasper-jager (approved):

PR #983, opened 2025-03-31
Warmup Django before exposing uWSGI worker

Yrob:
This way endusers will not experience a slow response when a new pod starts or a worker is restarted.
Based on http://uwsgi-docs.readthedocs.org/en/latest/articles/TheArtOfGracefulReloading.html#dealing-with-ultra-lazy-apps-like-django

Gets rid of slow responses during redeploys and scaling operations in kubernetes.


barrydaniels-nl (dismissed):
LGTM

barrydaniels-nl (approved):
LGTM2

PR #989, opened 2025-04-09
Jasper/bugfixes

jasper-jager:
Solve a few bugs:

-  Fix FieldError when trying to lookup on id instead of db_name

For temporal tables the identifier field was searched based on the id instead of the dbname. This causes issues when an id with an underscore is used, e.g. verblijfsobjectIdentificatie instead of verblijfsobject_identificatie. This fix now uses the field.db_name
instead of the id

- Catch valueerror on incorrect urls

When someone tried to access an url with {id} instead of an actual id a ValueError was raised. This fix catches the error and returns a 404.

- Fix bug where mainGeometry is not added as the default feature

When there are multiple geometries defined in the schema and the mainGeometry comes after another geometry, it was not added as the default feature. When the other geometry has required permission, this caused an improperlyconfigured error.

This fix makes sure the maingeometry is added as the default/first feature in wfs.

### DIFF
--- a/src/dso_api/dynamic_api/views/api.py
+++ b/src/dso_api/dynamic_api/views/api.py
@@ -141,7 +141,13 @@ class DynamicApiViewSet(DSOViewMixin, viewsets.ReadOnlyModelViewSet):
             # Allow fallback to old full id search (note this may need ?geldigOp=* to work)
             queryset = queryset.filter(models.Q(**{self.lookup_field: pk}) | models.Q(pk=pk))
         else:
-            queryset = queryset.filter(pk=pk)
+            try:
+                queryset = queryset.filter(pk=pk)
+            except ValueError as err:
+                raise NotFound(
+                    _("No %(verbose_name)s found matching the query")
+                    % {"verbose_name": queryset.model._meta.verbose_name}
+                ) from err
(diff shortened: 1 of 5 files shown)


barrydaniels-nl (approved):
LGTM

src/dso_api/dynamic_api/views/api.py
```diff
@@ -105,7 +105,7 @@ def lookup_field(self):
         """Overwritten from GenericAPIView to filter on temporal identifier instead."""
         if self.temporal.is_versioned:
             # Take the first field as grouper from ["identificatie", "volgnummer"]
-            return self.model.table_schema().identifier[0]
+            return self.model.table_schema().identifier_fields[0].db_name
```
vdboor, line 108: @jasper-jager zeker weten dat niet niet `.python_name` moet zijn? We gebruiken dat veld immers in het ORM..

jasper-jager: Goede, ik was eigenlijk niet zo bewust van de `.python_name`, maar de `.db_name` wordt mogelijk afgekort natuurlijk. Ik zal het nog even aanpassen

PR #1041, opened 2025-06-19
Fix versioned urls

pstokkink:
uses latest schema-tools with fix for nested dataset urls
fixes urls in versions on browsable api


jasper-jager (approved):
Let's go

